PR #1, opened 2016-09-23
Optional param to force-refresh a link

infojunkie:
As per https://mantis.meedan.com/view.php?id=5052

A few things to review:
- how to test that `refresh` is effective?
- tagged with FIXME a couple of FB tests (maybe for later)
- tagged with HACK the fact that a variable in the code is only used for tests - if that's OK, we can do the same for `refresh`.

### DIFF
--- a/app/controllers/api/v1/medias_controller.rb
+++ b/app/controllers/api/v1/medias_controller.rb
@@ -6,12 +6,13 @@ module Api
     class MediasController < Api::V1::BaseApiController
       include MediasDoc
       include MediasHelper
-      
+
       skip_before_filter :authenticate_from_token!, if: proc { request.format.html? || request.format.js? || request.format.oembed? }
       after_action :allow_iframe, only: :index
 
       def index
         @url = params[:url]
+        @refresh = params[:refresh] == '1'
         (render_parameters_missing and return) if @url.blank?
         @id = Digest::MD5.hexdigest(@url)
         render_timeout { @media = Media.new(url: @url) } and return
@@ -32,10 +33,10 @@ module Api
       def render_as_json
         @request = request
         begin
-          render_timeout { render_success('media', @media.as_json.merge({ embed_tag: embed_url })) and return }
+          render_timeout { render_success('media', @media.as_json({ force: @refresh }).merge({ embed_tag: embed_url })) and return }
         rescue Pender::ApiLimitReached => e
           render_error e.reset_in, 'API_LIMIT_REACHED', 429
-        rescue 
+        rescue
           render_error 'Could not parse this media', 'UNKNOWN'
         end
       end
@@ -53,9 +54,7 @@ module Api
 
       def render_as_html
         begin
-          @cache = true
           unless File.exist?(cache_path)
-            @cache = false
             save_cache
           end
           render text: File.read(cache_path), status: 200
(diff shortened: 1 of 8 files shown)


caiosba:
That's OK a variable that's used only for the tests. Regarding how to test the refresh, the best way is by using expectations. For example, when `refresh=1`, add this line as the first one of your test: `Media.any_instance.expects(:parse).twice`, then add two lines `Media.new(url: 'http://same-url.com').as_json(refresh: '1')`. For another test, do it like this: first line should be the expectation is that the `parse` method is called only once: `Media.any_instance.expects(:parse).once` and the same line twice: `Media.new(url: 'http://same-url.com').as_json(refresh: '0')`. Hope it's clear.

Once there are tests, please update the PR.

caiosba:
We could get rid of the `@cache` hack if we use expectations to test that too.

infojunkie:
Added these tests:
```
  test "should respond to refresh parameter" do
    Media.any_instance.expects(:parse).twice
    Media.new(url: 'https://twitter.com/caiosba').as_json(refresh: '1')
    Media.new(url: 'https://twitter.com/caiosba').as_json(refresh: '1')
  end

  test "should respond to refresh blank parameter" do
    Media.any_instance.expects(:parse).once
    Media.new(url: 'https://twitter.com/caiosba').as_json(refresh: '0')
    Media.new(url: 'https://twitter.com/caiosba').as_json
  end
```

but they succeed no matter how many `times()` I put in the expectation. I don't understand what's wrong.

caiosba:
Oh, what the `as_json` expects is `force: true`, not `refresh: '1'`. Since the `refresh` variable is set on the controller, it's better to test it at the controller level, to make sure the whole flow works (also, `as_json(force: true)` is from Rails, we don't need to test it). I would add a functional test that verifies whether the `parsed_at` attribute changes depending on the `refresh` parameter:

```ruby
  test "should ask to refresh cache" do
    authenticate_with_token
    get :index, url: 'https://twitter.com/caiosba/status/742779467521773568', refresh: '1', format: :json
    first_parsed_at = JSON.parse(@response.body)['data']['parsed_at']
    get :index, url: 'https://twitter.com/caiosba/status/742779467521773568', refresh: '1', format: :json
    second_parsed_at = JSON.parse(@response.body)['data']['parsed_at']
    assert second_parsed_at > first_parsed_at
  end 

  test "should not ask to refresh cache" do
    authenticate_with_token
    get :index, url: 'https://twitter.com/caiosba/status/742779467521773568', refresh: '0', format: :json
    first_parsed_at = JSON.parse(@response.body)['data']['parsed_at']
    get :index, url: 'https://twitter.com/caiosba/status/742779467521773568', format: :json
    second_parsed_at = JSON.parse(@response.body)['data']['parsed_at']
    assert_equal first_parsed_at, second_parsed_at
  end
```

infojunkie:
Added tests for the `refresh` param as per Caio's suggestion.

I also tried to add expectations to replace the `@cache` hack, but failed. Here's what I did:

```ruby
  test "should create cache file if does not exist" do
    Api::V1::MediasController.any_instance.expects(:save_cache).once
    get :index, url: 'http://twitter.com/caiosba', format: :html
    get :index, url: 'http://twitter.com/caiosba', format: :html
  end
```

However, the `expects` method seems to stub `save_cache` and does not call the original, making the test fail.

caiosba:
Yes, when you mock a method, it won't call the original method body anymore. For the `@cache` case, I suggest to mock the `as_json` method, because this way the cache is still saved. Something like this:

```ruby
test "should create cache file if does not exist" do
  Media.any_instance.expects(:as_json).returns({})
  get :index, url: 'http://twitter.com/caiosba', format: :html
  get :index, url: 'http://twitter.com/caiosba', format: :html
end
```

The `expects` method only passes if it's called once. So, if caching is not working, this test would fail. In order to verify that, try to comment this line: https://github.com/meedan/pender/blob/develop/app/controllers/api/v1/medias_controller.rb#L93

caiosba:
Also, I noticed that the test `should read from cache file if exists` is redundant... IMO we should remove it.

PR #2, opened 2016-10-05
Ticket #5050: Display error if the url is not valid

danielafeitosa:


app/controllers/api/v1/medias_controller.rb
```diff
@@ -14,8 +14,13 @@ def index
         @url = params[:url]
         @refresh = params[:refresh] == '1'
         (render_parameters_missing and return) if @url.blank?
+        (render_error('The URL is not valid', 'INVALID_VALUE') and return) unless @url =~ /\A#{URI::regexp(%w(http https))}\z/
         @id = Digest::MD5.hexdigest(@url)
-        render_timeout { @media = Media.new(url: @url) } and return
+        begin
+          render_timeout { @media = Media.new(url: @url) } and return
+        rescue Exception => e
+          render_error("Could not parse URL (#{e.message})", 'UNKNOWN') and return
```
caiosba: If this exception is thrown, it means that the URL validation can't handle all cases? I would like to avoid rescuing generic `Exception` class

app/controllers/api/v1/medias_controller.rb
```diff
@@ -14,8 +14,13 @@ def index
         @url = params[:url]
         @refresh = params[:refresh] == '1'
         (render_parameters_missing and return) if @url.blank?
+        (render_error('The URL is not valid', 'INVALID_VALUE') and return) unless @url =~ /\A#{URI::regexp(%w(http https))}\z/
```
danielafeitosa: Ok, I'll do it

app/controllers/api/v1/medias_controller.rb
infojunkie: Yes - plz review ticket #5174 for our approach to error handling

danielafeitosa: I created a new method to validate the url and now it rescues specific errors involving problems with invalid url (malformed or inexistent)

danielafeitosa:
@caiosba can you review it again, please?

PR #4, opened 2016-10-17
Ticket #5048: Parsing metatags from html pages

danielafeitosa:


test/models/media_test.rb
```diff
@@ -549,4 +549,36 @@ class MediaTest < ActiveSupport::TestCase
...
+    assert_equal 'http://hacktoon.com/static/img/facebook-image.png', d['picture']
+    assert_equal '', d['author_url']
+  end
+
+  test "should parse meta tags as fallback" do
+    m = create_media url: 'https://xkcd.com/1479'
+    d = m.as_json
+    assert_equal 'xkcd: Troubleshooting', d['title']
+    assert_equal 'xkcd: Troubleshooting', d['description']
+    assert_equal '', d['published_at']
```
caiosba: It would be good to have a case where title is different from description

PR #5, opened 2016-10-18
Feature/links fallbacks

danielafeitosa:
Set picture and author_url if they are not defined on page


danielafeitosa:
Yes, It includes the scheme, host and port

PR #9, opened 2016-10-24
Ticket #5250: Remove blank spaces from url beginning and end

danielafeitosa:


app/models/media.rb
```diff
@@ -96,6 +96,7 @@ def normalize_url
   end
 
   def follow_redirections
+    self.url = self.url.strip
```
infojunkie: This command should be in the normalization function, not in the redirection function.

danielafeitosa: With the get_canonical_url, the normalization function is the third step when creating a new media. The first one is follow_redirections, then get_canonical_url and then normalization. The url already needs to be valid (without blank spaces on left and right) in the first step or it will crash when Pender try to parse it

PR #20, opened 2016-11-03
Feature/fb live post

danielafeitosa:


infojunkie:
Do we need to handle removed posts per provider? Feels like we should have a general mechanism for handling removed posts.

danielafeitosa:
Karim, Facebook is not returning a `404` error code. It is returning a `200`. We only know it is not a valid one when trying to parse it.

PR #23, opened 2016-11-08
Ticket #5425: Return error message when url is not valid

danielafeitosa:


infojunkie:
I have missed this PR. Is it still relevant?

danielafeitosa:
It is still relevant, but I think we can wait until we define how to handle the different cases of parsing errors

PR #31, opened 2016-12-22
Ticket #5552: Fix parsing of url when raises a encoding header problem

danielafeitosa:


app/models/media.rb
```diff
@@ -237,7 +237,7 @@ def absolute_url(path = '')
     if path =~ /^https?:/
       path
     elsif path =~ /^\/\//
-      self.parse_url(self.url).scheme + path
+      "#{self.parse_url(self.url).scheme}:#{path}"
```
infojunkie: Why change this line to use string interpolation and not the line below?

danielafeitosa: The interpolation was just to add the ":".. But I'll change it to be similar to the other line

PR #34, opened 2017-01-13
Ticket #5557: Add support to dropbox links

danielafeitosa:


app/models/concerns/media_dropbox_item.rb
```diff
@@ -0,0 +1,34 @@
+module MediaDropboxItem
+  extend ActiveSupport::Concern
+
+  DROPBOX_URL = /^https?:\/\/(www\.)?dropbox\.com\/s\/([^\/]+)/
```
caiosba: What if the input URL is already `dl.dropbox.com/...`?

app/models/concerns/media_dropbox_item.rb
```diff
@@ -0,0 +1,34 @@
...
+  end
+
+  def parse_from_dropbox_html
+    self.data['description'] = self.doc.at_css("meta[property='og:description']").attr('content').strip
+    self.data['title'] = self.doc.at_css("meta[property='og:title']").attr('content').strip
+    self.data['picture'] = self.doc.at_css("meta[property='og:image']").attr('content')
+    self.data['author_url'] = top_url(self.url)
+  end
+
+  def html_for_dropbox_item
```
infojunkie: Does Dropbox offer an embeddable view? It would be better to use their embed rather than a link tag.

app/models/concerns/media_dropbox_item.rb
danielafeitosa: It will raise an error.. I'll fix it

PR #37, opened 2017-03-02
Ticket #5815: Handle exception when raises error when try to get oemb…

danielafeitosa:
…ed data


infojunkie:
Looks good to me although I don't know what `self.absolute_url` does :-)

danielafeitosa:
This method (https://github.com/meedan/pender/blob/develop/app/models/media.rb#L249-L258) make sure that the path/url has a scheme and a domain

PR #38, opened 2017-03-03
Ticket #5757: Get username (slug) and name from medias

danielafeitosa:
Get username (slug) from Facebook user profiles and pages
Get author_name from Instagram

### DIFF
--- a/app/models/concerns/media_facebook_item.rb
+++ b/app/models/concerns/media_facebook_item.rb
@@ -192,6 +192,11 @@ module MediaFacebookItem
     <div class="fb-post" data-href="' + self.url + '"></div>'
   end
 
+  def get_facebook_slug_from_html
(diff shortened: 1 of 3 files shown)


app/models/concerns/media_facebook_item.rb
```diff
@@ -200,7 +205,7 @@ def data_from_facebook_item
       self.data['text'].strip!
       self.data['media_count'] = 1 unless self.url.match(/photo\.php/).nil?
       self.data.merge!({
-        username: self.data['user_name'],
+        username: self.get_facebook_slug_from_html || self.data['user_name'],
```
amoedoamorim: Perhaps this should return an empty string instead of falling back to the real name when username is unknown, given we already have a dedicated field for the user's real name?

PR #45, opened 2017-06-06
Ticket #6069: Removed `bio` field from list of facebook user_fields

danielafeitosa:
The field `bio` is deprecated for versions v2.8 and higher and the
content will be available on `about` field


caiosba:
You just removed from `facebook_user_fields`, but I think you need to remove from `facebook_page_fields` too.

caiosba:
We already fallback to the `about` field when the `bio` field is not available: https://github.com/meedan/pender/blob/develop/app/models/concerns/media_facebook_profile.rb#L75. Now that the `bio` field will never be available, we can remove it from here too.

danielafeitosa:
The `bio` is still available for `page` objects :)

PR #60, opened 2017-08-29
Ticket #6173: Add support to pages that redirect from https to http

danielafeitosa:
The redirection `https` to `http` is allowed only when the request
doesn't have credentials for `http_basic_authentication` header

Also:
- search for `author_name` in more meta tags
- decode url when trying to parse FB username;

### DIFF
--- a/app/models/concerns/media_page_item.rb
+++ b/app/models/concerns/media_page_item.rb
@@ -7,7 +7,7 @@ module MediaPageItem
 
   def data_from_page_item
     if self.doc.nil?
-      self.doc = self.get_html
+      self.doc = self.get_html({ allow_redirections: :all })
       self.data['raw']['metatags'] = get_metatags(self)
     end
     handle_exceptions(self, RuntimeError) do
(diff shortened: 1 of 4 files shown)


app/models/concerns/media_page_item.rb
```diff
@@ -7,7 +7,7 @@ module MediaPageItem
 
   def data_from_page_item
     if self.doc.nil?
-      self.doc = self.get_html
+      self.doc = self.get_html({ allow_redirections: :all })
```
infojunkie: can we be caught in an infinite redirect loop here?

danielafeitosa: The `allow_redirections` only will allow redirections http => https and https => http. So, I think that we can be caught in an infinite redirect loop in the case where the server is badly configured. I can try to reproduce this case on test

danielafeitosa: OpenUri will raise an error if there is a redirection loop: https://apidock.com/ruby/v1_9_3_392/OpenURI/open_loop/class

PR #74, opened 2017-11-11
Fix YT username #6519

infojunkie:


test/models/media_test.rb
```diff
@@ -17,7 +17,7 @@ class MediaTest < ActiveSupport::TestCase
     assert_equal 'Porta dos Fundos', data['title']
     assert_equal 'portadosfundos', data['username']
     assert_equal 'Porta dos Fundos', data['author_name']
-    assert_equal 'user', data['subtype']
+    assert_equal 'channel', data['subtype']
```
caiosba: I didn't understand why the `subtype` is returning `channel`. The URL is https://www.youtube.com/user/portadosfundos and I verified that YouTube doesn't redirect to https://www.youtube.com/channel/portadosfundos. And the `subtype` is extracted from the URL.

app/models/concerns/media_youtube_profile.rb
```diff
@@ -90,8 +90,8 @@ def data_from_youtube_profile
 
   def get_youtube_username
     match = self.url.match(/^https?:\/\/(www\.)?youtube\.com\/user\/([^\/]+)/)
-    username = match.nil? ? self.data['title'].gsub(/[^a-zA-Z0-9]/, '') : match[2]
-    username
+    username = match.nil? ? self.data[:raw][:api][:title].gsub(/[^a-zA-Z0-9]/, '') : match[2]
```
caiosba: `self.data['title']` is equal to `self.data[:raw][:api][:title]` (as per https://github.com/meedan/pender/blob/develop/app/models/concerns/media_youtube_profile.rb#L30). What's different here?

infojunkie: Afaict, the difference is that the function is called before `data[title]` is set.

caiosba: ok

test/models/media_test.rb
infojunkie: Yes, need to dig deeper into this one.

infojunkie: Noting that the Pender API returns a url of `https://www.youtube.com/channel/UCEWHPFNilsT0IfQfutVzsag` - I expect this is the canonical URL that is sniffed by Pender on the profile page, and the `subtype` is extracted afterwards. I am not even sure what `subtype` is supposd to represent :-)

caiosba: The idea was to differentiate between "people" and "pages"... but I don't think it has any practical implications :)

infojunkie: OK in that case I am merging :-)

PR #89, opened 2018-04-04
Ticket #5136: Avoid Facebook API

caiosba:
Do not use Facebook API anymore to parse Facebook profiles or posts. Instead, use only HTML parsing.

### DIFF
--- a/app/models/concerns/media_facebook_item.rb
+++ b/app/models/concerns/media_facebook_item.rb
@@ -81,6 +81,7 @@ module MediaFacebookItem
 
   def parse_from_facebook_html
     self.doc = self.get_html(html_options)
+    return if self.doc.nil?
     self.get_facebook_info_from_metadata
     self.get_facebook_text_from_html
     self.get_facebook_owner_name_from_html
@@ -202,11 +203,12 @@ module MediaFacebookItem
       self.parse_from_facebook_html
       self.data['text'].strip! if self.data['text']
       self.data['media_count'] = 1 unless self.url.match(/photo\.php/).nil?
+      self.data['author_name'] = 'Not Identified' if self.data['author_name'].blank?
       self.data.merge!({
         username: self.get_facebook_username || self.data['author_name'],
         title: self.data['author_name'] + ' on Facebook',
         description: self.data['text'] || self.data['description'],
-        picture: self.data['photos'].first,
+        picture: self.data['photos'].nil? ? '' : self.data['photos'].first,
         html: self.html_for_facebook_post,
         author_name: self.data['author_name'],
         author_url: 'http://facebook.com/' + self.data['user_uuid'].to_s
--- a/app/models/concerns/media_facebook_item.rb
+++ b/app/models/concerns/media_facebook_item.rb
@@ -92,10 +92,23 @@ module MediaFacebookItem
 
   def get_facebook_info_from_metadata
     get_metatags(self)
-    metadata = self.get_opengraph_metadata
-    self.data['author_name'] = metadata['title']
-    self.data['text'] = metadata['description']
-    self.data['photos'] = metadata['picture'].nil? ? [] : [metadata['picture']]
+    metadata = self.get_opengraph_metadata || {}
+    self.data['metadata'] = metadata
+    self.data['author_name'] = metadata['title'].nil? ? self.get_facebook_title_from_html : metadata['title']
+    self.data['text'] = metadata['description'].nil? ? self.get_facebook_description_from_html : metadata['description']
+    self.data['photos'] = metadata['picture'].nil? ? self.get_facebook_photos_from_html : [metadata['picture']]
+  end
+
+  def get_facebook_description_from_html
+    self.doc.css('span[data-testid="event-permalink-details"]').text unless self.url.match(EVENT_URL).nil?
+  end
+
+  def get_facebook_title_from_html
+    self.doc.css('#pageTitle').text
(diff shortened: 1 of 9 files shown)


app/models/concerns/media_facebook_item.rb
```diff
@@ -250,24 +207,28 @@ def data_from_facebook_item
     self.screenshot_script = 'function closeLoginModal(){if(document.getElementById("headerArea")){document.getElementById("headerArea").style.display="none";}else{setTimeout(closeLoginModal,1000);}}closeLoginModal();'
     handle_exceptions(self, StandardError) do
       self.parse_facebook_uuid
-      self.parse_from_facebook_html unless self.parse_from_facebook_api
+      self.parse_from_facebook_html
       self.data['text'].strip! if self.data['text']
       self.data['media_count'] = 1 unless self.url.match(/photo\.php/).nil?
+      self.data['author_name'] = 'Not Identified' if self.data['author_name'].blank?
```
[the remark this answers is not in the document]
caiosba: yes, agree... at this point I just kept the same behavior as before (`Not Identified` for `author_name` and empty strings for everything else)

app/models/concerns/media_facebook_profile.rb
```diff
@@ -11,93 +11,116 @@ module MediaFacebookProfile
...
+  def parse_facebook_page
+    html = self.get_facebook_profile_html
+    page = self.get_facebook_profile_page
+    match = html.match(/"name":"([^"]+)","pageID":"([^"]+)","username":([^,]+),"usernameEditDialogProfilePictureURI":"([^"]+)"/)
+    json = JSON.parse('{' + match[0] + '}')
+    
+    data = {}
+    data['name'] = json['name']
+    data['username'] = json['username']
+    data['description'] = page.css('meta[name=description]').first.attr('content').gsub(/.*talking about this. /, '')
```
infojunkie: are we ensuring English language here?

caiosba: yes, I set the header `Accept-Language: en` in the crawler request

app/models/concerns/media_facebook_profile.rb
```diff
@@ -11,93 +11,116 @@ module MediaFacebookProfile
...
     end
-    self.get_facebook_likes
+    @page
+  end
+
+  def parse_facebook_user
+    page = self.get_facebook_profile_page
+
+    data = {}
+    data['name'] = page.css('#fb-timeline-cover-name').first.text
```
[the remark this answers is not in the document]
caiosba: it will fail and we'll receive a notification on Errbit

app/models/concerns/media_facebook_profile.rb
infojunkie: looks like we are below, yes

app/models/concerns/media_facebook_item.rb
```diff
@@ -175,10 +93,23 @@ def parse_from_facebook_html
 
   def get_facebook_info_from_metadata
     get_metatags(self)
-    metadata = self.get_opengraph_metadata
-    self.data['author_name'] = metadata['title']
-    self.data['text'] = metadata['description']
-    self.data['photos'] = metadata['picture'].nil? ? [] : [metadata['picture']]
+    metadata = self.get_opengraph_metadata || {}
```
infojunkie: isn't there another parser for OG tags? do we need to replicate its work here?

caiosba: I'm reusing the same OG parser we use for the generic page parser: https://github.com/meedan/pender/blob/develop/app/models/concerns/media_page_item.rb#L48

infojunkie: good - I meant: is the generic page parser always invoked even for Fb links? if so, we should not need to re-invoke it here.

caiosba: no it's not... I'm just reusing the method but it's called only once (which is here at this line)

infojunkie: weird... I thought we had some parsers that applied to all links.. let's discuss too

caiosba: no, we have steps that apply to all links (for example, archiving, extract metatags (which are reused for the OG extracting), extract JSON-LD, extract Schema.org tags, etc.)

infojunkie: that's what I was asking: if extract metatags is always called, do we need to call it again in the Fb step?

caiosba: we call it here just to map the metatags to field values, but we extract only once

infojunkie: I see thanks

app/models/concerns/media_facebook_profile.rb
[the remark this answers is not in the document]
infojunkie: that's good. if I read correctly, this is a consequence of the `.css()` call returning `nil` and us trying to call `.first` on whatever is returned by it.. will this happen for every other call in here that tries to read from the DOM ? we need to make sure that a class name or id that is not found always generates an exception - unless it is optional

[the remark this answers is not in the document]
caiosba: I tried to do that, yes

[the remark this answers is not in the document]
infojunkie: great thx

PR #164, opened 2019-12-11
471 hosam config runtime

huslage:


huslage:
The `docker-entrypoint.sh` script is for local development and doesn't need the configurator to be applied inside the container.

PR #223, opened 2020-08-20
Ticket #8185: add new tests

danielevalverde:


test/models/media_test.rb
```diff
@@ -1032,4 +1033,14 @@ class HttpRedirectionLoop < StandardError
     assert_equal api_key.settings[:config][:storage], Pender::Store.current.instance_variable_get(:@storage)
   end
 
+  test "should not get metrics from Facebook when facebook api key is empty" do
+    url = 'http://www.facebook.com/people/\u091C\u0941\u0928\u0948\u0926-\u0905\u0939\u092E\u0926/100014835514496'
+    api_key = create_api_key application_settings: { config: { facebook:{ app_id: '', app_secret: '' } }}
+    response = Media.get_metrics_from_facebook(url, api_key.id, 10)
```
danielafeitosa, line 1039: To get the fb metrics this method calls `request_metrics_from_facebook` (that uses the fb config you are testing) and it is stubbed: https://github.com/meedan/pender/blob/95a45153fb96e85d52321f1e372d5b443d8e57b7/test/test_helper.rb#L66
Even if you send the correct config the metrics won't be retrieved

test/models/media_test.rb
```diff
@@ -1032,4 +1033,14 @@ class HttpRedirectionLoop < StandardError
...
+    assert_nil response
```
danielafeitosa, line 1040: `get_metrics_from_facebook` is expected to return `nil` when the last argument is 10 or higher.
You can verify this sending number 1. In this case, the method will return the delayed job id: https://github.com/meedan/pender/blob/develop/app/models/concerns/media_facebook_engagement_metrics.rb#L36

test/models/media_test.rb
```diff
@@ -1032,4 +1033,14 @@ class HttpRedirectionLoop < StandardError
...
+    m = create_media url: url
+    data = m.as_json
+    assert_nil data['metrics']['facebook']
```
danielafeitosa, line 1043: If there's a problem with the metrics request, Pender expect it to be an empty hash (`{}`), not `nil`
https://github.com/meedan/pender/blob/develop/app/models/concerns/media_facebook_engagement_metrics.rb#L31-L34

PR #236, opened 2020-12-16
Normalize SSM parameter names to lower case.

sonoransun:
This is one of a couple pull requests to normalize SSM parameter prefixes to lowercase. This better matches the format in use in Terraform and AWS.

Live and QA configuration has been updated as per the [SSM import branch of configurator](https://github.com/meedan/configurator/tree/ssm-migration/ssm_import).

This change can be merged when convenient; the existing configuration lives along side the new format to ensure a seamless migration.  The deprecated settings can be deleted once the migration is complete. (This will be a separate sysops ticket.)

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -39,10 +39,10 @@ deploy_qa:
     - pip install docutils==0.16
     - pip install awscli==1.18.188
     - pip install ecs-deploy==1.11.0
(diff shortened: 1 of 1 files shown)


caiosba:
Do we still need this PR, @sonoransun ?

sonoransun:
Yes!  We have both parameter sets (Upper and Lower case) available, but this brings pender in line with the rest of the SSM migration.

PR #239, opened 2021-04-02
Remove deprecated params

sonoransun:
During a review of SSM parameters in Live, it was observed that the pender service is still referencing some long deprecated parameters.

This change updates the ECS deploy commands to use exclusive secrets and environment parameters. Once deployed, the pender tasks will no longer reference deprecated parameters like the `s3_*` values and the UpperCase `/Live/` prefix.


caiosba (approved):
My review was basically: the qa section just mentions "qa", the live section just mentions "live", and they have the same number of occurrences :)

PR #243, opened 2021-05-20
Test/436 Run tests without Configurator

danielevalverde:


test/controllers/medias_controller_test.rb
```diff
@@ -606,9 +618,9 @@ def setup
     assert_response :success
     sleep 2
     data1 = Pender::Store.current.read(id1, :json)
-    assert_match /The Checklist: How Google Fights #Disinformation/, data1['title']
+    assert_not_nil data1['title']
```
danielafeitosa, line 621: Could you change this to verify it's not blank instead of `nil`, please?
We want to make sure the `title` is never blank

PR #244, opened 2021-05-21
use Github Actions for tests and standardize app startup

dmou:
High-level changes:
- Tests are run, without configurator, in Github Actions and use /test/pender/* values from AWS SSM
- Tests are run on every commit and #check-dev is notified on every result
- Travis is removed entirely
- Dockerfile is unified for local/dev and deployments- but deployments are still using production/Dockerfile for now
- A Makefile is the primary method of interaction with the containers now, more on that below
- The docker-compose.yml file no longer mounts pender/ to /app/ in the container. The container must be rebuilt for any and all file updates


caiosba:
Can we close this now, @dmou ?

dmou:
> Can we close this now, @dmou ?

hmm it has some stuff for dockerfile consolidation :thinking: but yes let's delete it, if I need it later I'll just use my local copy or archives

PR #273, opened 2022-08-09
Set TikTok parsing defaults

hartsick:
Set whatever data we can when parsing fails for both TikTok
items and profiles. Also adds some tests to reenforce behavior.

Lastly, starts to explicitly mock server responses and enforces that 
we're not making non-real external calls in tests when we don't need to.
Hoping we can slowly expand this pattern and speed up our tests!

### DIFF
--- a/app/models/media.rb
+++ b/app/models/media.rb
@@ -75,6 +75,7 @@ class Media
     if options.delete(:force) || Pender::Store.current.read(Media.get_id(self.original_url), :json).nil?
       handle_exceptions(self, StandardError) { self.parse }
       self.data['title'] = self.url if self.data['title'].blank?
+      self.data['description'] = self.url if self.data['description'].blank?
       data = self.data.merge(Media.required_fields(self)).with_indifferent_access
       Pender::Store.current.write(Media.get_id(self.original_url), :json, cleanup_data_encoding(data))
       self.upload_images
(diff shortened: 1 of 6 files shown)


test/models/tiktok_test.rb
```diff
@@ -41,6 +41,30 @@ class TiktokTest < ActiveSupport::TestCase
...
+    assert_equal '7090122043793984795', data['external_id']
+    assert_match 'https://www.tiktok.com/@ken28gallardo', data['author_url']
+    assert_nil data['error']
+    assert_equal '@ken28gallardo', data['username']
+  end
+end
+
+class TikTokUnitTest < ActiveSupport::TestCase
+  def setup
+    WebMock.enable!
```
hartsick, line 59: I'd like to move some of these request stubs up to the test_helper level eventually, but will take a bit of time - going to start trying to isolate tests as I touch the files.

app/models/media.rb
```diff
@@ -75,6 +75,7 @@ def as_json(options = {})
     if options.delete(:force) || Pender::Store.current.read(Media.get_id(self.original_url), :json).nil?
       handle_exceptions(self, StandardError) { self.parse }
       self.data['title'] = self.url if self.data['title'].blank?
+      self.data['description'] = self.url if self.data['description'].blank?
```
hartsick, line 78: Since it seems like we've been asking for the description to be set to the URL by default as well as title, I went ahead and made the change here, which will impact all parsers.

caiosba: Makes sense, thanks!

test/models/tiktok_test.rb
caiosba: Looks good. Maybe eventually we can split those classes into different test files and then yes, that common logic could be in a `setup` method used by both.

caiosba (approved):

PR #275, opened 2022-08-10
Immediately send metrics collection to background job

hartsick:
Previously we were making our initial request for metrics
synchronously, then scheduling background jobs for follow-up
updates. This slowed down the in-browser response (since we
had to wait for multiple metrics-related network requests) and
also sometimes errored (if we ran into issues fetching metrics).

Instead, this PR moves all metric checking to background.

It also splits out metrics-related tests into their own files so
that we can more easily find them.


app/models/concerns/media_facebook_engagement_metrics.rb
```diff
@@ -9,14 +9,28 @@ module MediaFacebookEngagementMetrics
...
-    self.class.get_metrics_from_facebook(self.original_url, ApiKey.current&.id, 0, facebook_id)
+    MetricsWorker.perform_async(self.original_url, ApiKey.current&.id, 0, facebook_id)
   end
 
   def is_a_facebook_post?
     self.data && self.data['provider'] == 'facebook' && self.data['type'] == 'item'
   end
 
   module ClassMethods
+    def get_metrics_from_facebook(url, key_id, count = 0, facebook_id = nil)
```
hartsick, line 20: Just moving this method up so that it reads a little clearer, since this method is the main entrypoint

PR #276, opened 2022-08-10
Remove dependency on inotify

hartsick:
inotify causes problems on M1 macs using the linux/x86_64 emulator in Docker
(see https://github.com/docker/for-mac/issues/5321).

We were only using this to boot up Sidekiq in the pender-background service,
and the approach we were using was both not recommended by Sidekiq 6.0, and
causing a segfault when building the Docker image using arm architecture. Here
is the error that happened when the container started in `arm`:
```
check-pender-background-1  | ERROR: PID file creation was removed in Sidekiq 6.0, please use a proper process supervisor to start and manage your services
check-pender-background-1  | ERROR: Logfile redirection was removed in Sidekiq 6.0, Sidekiq will only log to STDOUT
check-pender-background-1  | ERROR: Daemonization mode was removed in Sidekiq 6.0, please use a proper process supervisor to start and manage your services
```

This commit gets pender-background processing jobs locally on M1 Macs using the emulator,
though it is a slightly different approach than is present in api-background
(see https://github.com/meedan/check-api/commit/1abe89d18de4a1e38c5b7d47dcb8316fb2fda4c0). 
@huslage do you happen to have context on why we had the environment-specific
logic in booting up sidekiq in the link above and if we need to try to keep that behavior
in the pender background job?


hartsick:
Merging this in so that I can get it on develop - will add note in Jira about need for restart

PR #281, opened 2022-08-16
Check 2208 convert provider modules in pender to classes

hartsick:
Opening for visibility; planning to keep this open as I develop. Thoughts welcome!

To do:
- [x] Extract Metrics and isolate tests
- [x] Refactor and add unit tests for KwaiItem
- [x] Refactor and add all tests for DropboxItem
- [x] Refactor TiktokProfile
- [x] Refactor TiktokItem
- [x] Refactor InstagramProfile
- [x] Refactor InstagramItem
- [x] Refactor TwitterProfile
- [x] Refactor TwitterItem
- [x] Refactor YoutubeProfile
- [x] Refactor YouTubeItem
- [x] Refactor FacebookProfile
- [x] Refactor FacebookItem
    - [x] ~Unavailable page behavior (don't update URL, don't show embed)~ think I've managed this; want to verify on QA
    - [x] ~Shared content~ going to deploy without; want to verify on QA
    - [x] Handling of events
    - [x] Parse data from metatags
    - [x] Parse data from jsonld
- [x] Refactor PageItem
    - [x] Get Twitter information
- [x] Refactor OembedItem
- [x] Assign url back to Media object
- [x] At end: remove unused functions on Media
- [x] At end: remove unused functions on MediasHelper class
- [x] Check tests in media and controller
- [x] Before end: Make sure all skipped tests are unskipped / dealt with
    - [x] Oembed tests
- [x] Before end: Verify fallback behavior is as we want it (eg all non-default keys are empty, rather than returning partial or incorrect data)
- [x] Before end: Check that hash merging behavior for data working as expected when returned to Media
- [x] Before end: Update documentation
- [ ] In future: Differentiate private and public methods on RequestHelper
- [ ] In future?: Standardize error handling and reporting for parsers
- [ ] In future?: Reverse test setup/teardown functionality (eg, default to isolation and explicitly set non-isolated integration tests)
- [ ] In future?: Remove Sidekiq inline from test setup to prevent running integration requests 10x times
- [x] In future?: Set metatags and json LD content as part of parser initalization

Some notes:
- Twitter-related things return an error from the Pender service when Twitter rate limits us (Pender::ApiLimitReached). This is different behavior than when we believe we're rate limited by Instagram and TikTok. Should we have the same behavior and, if so, which?
- ~In TwitterProfile, we set a bunch of picture-related fields on the data object that we don't do anywhere else ([:picture][:normal][:bigger][:mini][:original]). We don't use them elsewhere in Pender—just use original to set the same fields as other parsers. Do we use them in other applications, or can we remove to standardize?~ Decided to remove!
- In YoutubeProfile and Item we aren't assigning a top level error when we get errors from the API. Should we? I began sending errors to Errbit. We also are only handling a few Youtube exceptions, which is similar to Twitter but different than other providers.
- We have a few view partials, including one that render html (custom.html.erb) - what do we use them for? Can we delete?


hartsick:
Integration tests running on this branch: https://github.com/meedan/check-web/tree/CHECK-2208-convert-provider-modules-in-pender-to-classes

PR #287, opened 2022-09-21
Omit configurator

sonoransun:
This change removes use of the configurator repo for application configuration. Environment variables set via SSM are now used for QA and Live deployments.


production/bin/create_configs.sh
```diff
@@ -0,0 +1,76 @@
...
+
+# generate configurartion files with group and world read permissions.
+umask 022
+
+echo "Starting application configuration. Processing ENV settings."
+
+# Move default configs into place.
+# For most environments, these settings are overridden in ENV set from SSM.
+(
+  cd config
```
hartsick, line 11: If we are only using this config generation script in production, I wonder if we should be copying over many of these files? I worry about using development-related configuration (like in config.yml, since we have default values for all environments; or cookies.txt) accidentally rather than bailing on the deploy.

production/bin/create_configs.sh
```diff
@@ -0,0 +1,76 @@
...
+    else
+      echo "Using decoded database config from ENV var: ${database_config} ."
+      mv $WORKTMP database.yml
+      sha1sum database.yml
+    fi
+  fi
+
+  # Populate production environment config from SSM:
+  WORKTMP=$(mktemp)
+  if [[ -z ${environments_production+x} ]]; then
```
hartsick, line 60: Could you help me understand how this works? I want to make sure that these values override the default ones given in config.yml, since right now with the file copied above those would all be development values. Where do variables like `environments_production` come from?

hartsick:
Hi @sonoransun - excited for this! Left a couple of questions on parts that I didn't understand.

production/bin/create_configs.sh
sonoransun: My understanding is that we'll be using the default configs for unit tests and integration tests. In these situations, we want to use the default configs as they are sufficient for testing, and we need to move them into place if they don't exist for the application to start under test.

The concern about starting with a default configuration in Live is indeed a concern - typically when I make changes like this I will watch the deploy in ECS Events and follow the CloudWatch logs to confirm things start as expected, and configs are updated from SSM as intended. Since we do this migration once (moving from configurator to SSM driven configuration) I think the risk is manageable, but maybe we can make this more robust like you mention, where Live deploys fail rather than use defaults.

production/bin/create_configs.sh
sonoransun: I realized that a confluence document describing the SSM driven configuration of Check applications would be useful here. Let me get this posted and point at the relevant parts in the doc. (Basically, the SSM parameter of this name, is mapped into the task ENV with same name, and this base64 value is used to generate the config file...)

production/bin/create_configs.sh
hartsick: If we can confirm that everything is working well to start then that sounds good for the migration, especially since we'll want this for the cookies.txt changes that are already on develop and about to be deployed!

I'd like to get these changes made if we can, though, since I worry about someone forgetting to set a new config value going forward and defaulting to a dev/test value. We could have copying example files be run as part of the test run on CI but not a part of the start step, I think.

production/bin/create_configs.sh
hartsick: Super helpful, thanks!

hartsick (approved):

production/bin/create_configs.sh
```diff
@@ -0,0 +1,95 @@
...
+  fi
+
+  # As mentioned above, if we are missing required configs in QA or Live environemnts
+  # emit an error detailing the omitted configurations and exit with error.
+  # This will promptly fail an attempted deployment, rather than deferring to
+  # application restarts to indicate error.
+  if [[ "$DEPLOY_ENV" == "qa" || "$DEPLOY_ENV" == "live" ]]; then
+    if [[ "$missing_configs" != "" ]]; then
+      echo "Error: missing required configurations: $missing_configs exiting."
+      kill 0
```
sonoransun, line 89: One of the hiccups here is that the configuration process is done in a subshell inside the `config/` directory. In order to properly propagate up the exit error, we use a `kill 0` idiom here, which is a bit unconventional but perfectly effective.

PR #289, opened 2022-09-23
Send initial data to Honeycomb

hartsick:


config/config.yml.example
```diff
@@ -117,6 +117,17 @@ development: &default
   #
   perma_cc_key: # '<PERMA.CC ACCESS KEY>'
 
+  # Open Telemetry configuration, for reporting to Honeycomb
+  # See initializers/open_telemetry.rb for usage.
+  #
+  # OPTIONAL
+  #
+  otel_exporter_otlp_endpoint: "https://api.honeycomb.io"
+  otel_exporter_otlp_headers: "x-honeycomb-team=<DEV API KEY>"
```
hartsick, line 126: Right now, while we're using Honeycomb Classic, we also need to include the Dataset name in the header, eg:
```
x-honeycomb-team=<DEV API KEY>,x-honeycomb-dataset=check-qa
```
Intending to change this (and update local API key locally) before we merge in.

PR #298, opened 2022-10-11
Always set a normalized URL

hartsick:
Previously we were getting different URLs based on when we were setting the URL. We had updated Media initialization, but hadn't updated the other places after Media is built that we might update the URL. This commit overrides the existing setter method so that we make sure to always normalize the URL going forward.


hartsick:
Closing in favor of fixing this on the Check API side and keeping our URL normalization as-is

PR #309, opened 2022-11-23
Re-queue webhook notification job when webhook returns an errored response

hartsick:
This PR does a few things:
1. Adds a delay before kicking off the archiving jobs, to give Check API a headstart to start creating the link, project media, and annotation that will be needed to associate the archival information
2. Begins raising exception when the webhook notification to Check API returns an unsuccessful response, which Sidekiq will then retry.
3. Uses WebMock to stub external requests to the endpoint in tests, and removes dependency on the external webhook endpoint on Caio's site (partly because it was returning 301, which is considered unsuccessful)

**Given that we should always expect the Check webhook to be working when we send a webhook notification from Pender, I think always retrying when the endpoint errors makes sense regardless of what that error returned from Check API is, but I'd love feedback on that**. Right now the webhook endpoint will error in the case of an incorrect bot name (404) or an incorrect secret token (400), and I'll be adding the race conditions (link, project media, or archiver annotation unavailable; 425).

### DIFF
--- a/app/models/media.rb
+++ b/app/models/media.rb
@@ -133,7 +133,6 @@ class Media
     data = Pender::Store.current.read(id, :json)
     unless data.blank?
       newdata.each { |key, value| data[key] = data[key].is_a?(Hash) ? data[key].merge(value) : value }
-      data['webhook_called'] = @webhook_called ? 1 : 0
       Pender::Store.current.write(id, :json, data)
     end
     data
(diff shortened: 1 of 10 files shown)


app/models/media.rb
```diff
@@ -150,17 +150,21 @@ def self.notify_webhook(type, url, data, settings)
...
         @webhook_called = true
+        # .value raises an exception if the response is unsuccessful, but otherwise
+        # returns nil (terribly named method, imo). This is a way to raise that
+        # exception if request failed and still return a successful response if present.
+        response.value || response
       rescue StandardError => e
         PenderAirbrake.notify(e, url: url, type: type, webhook_url: settings['webhook_url'])
         Rails.logger.warn level: 'WARN', message: 'Failed to notify webhook', url: url, type: type, error_class: e.class, error_message: e.message, webhook_url: settings['webhook_url']
-        return false
+        raise(e)
```
caiosba, line 162: Will this trigger a report to Errbit? If so, what do you think about deleting line 160?

caiosba (approved):
I just added one question, but looks great overall!

app/models/media.rb
hartsick: I was wondering the same thing, but wasn't sure and haven't checked locally yet. Going to see now.

hartsick: I think what I actually should do here is make our new exception classes related to things happening to early descend from `Pender::RetryLater` so that they're automatically ignored when the exception is raised, since it will be retried in the future. I'll also delete the Airbrake reporting line on 160!

caiosba: Yes, that sounds great, thanks!

PR #318, opened 2022-12-15
Prevent exponential creation of metrics jobs on failure

hartsick:
Previously, in some cases we were exponentially creating new Metrics updates for URLs beacuse we queued the next update inside the existing background job, and relied on a passed variable to dictate when to stop scheduling jobs (that on error would not always be set).

This instead schedules all of our updates at the start (for ten days, the previous setting) and caps the amount of retries up to one day (according to the Sidekiq documentation for back-off timing). This means there will be a max of 130 metrics fetches per link. We could reduce this if we wanted (either by turning off retries entirely, or - maybe better - only retrying the first, since it's presumably most important). Curious for your thoughts @caiosba.

In theory we could stop passing count / update_number, but this would change the signature of the method, which would cause issues for currently queued Sidekiq jobs, so I just left it and used it for some logging.

### DIFF
--- a/app/models/media.rb
+++ b/app/models/media.rb
@@ -83,7 +83,7 @@ class Media
       self.upload_images
     end
     self.archive(options.delete(:archivers))
-    Metrics.get_metrics_from_facebook_in_background(self.data, self.url, ApiKey.current&.id)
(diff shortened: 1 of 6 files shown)


caiosba (approved):
Looks good! If we still have issues, I think we can just disable Sidekiq retry entirely and just keep the 10 scheduled jobs.

PR #342, opened 2023-03-22
Change out Sentry for Airbrake / Errbit

hartsick:
We recently got access to a Sentry account, so want to move our error reporting over to the service.

This commit replaces Airbrake/Errbit with Sentry, and updates our method that we use to send information to our exception service. It also sets up Sidekiq's death_handlers to send information to Sentry when we pass the maximum number of retries for a job.

If sentry_dsn is unset then the application will not report to Sentry. This is configured unset by default in the dev and test environments, but can be optionally configured to send if we want to test exception reporting locally. This should also help us catch any bad interactions between OpenTelemetry and Sentry in dev and test environements (since the libraries are always loaded).

CV2-2897


hartsick:
This is failing for an unrelated reason, captured in another ticket. We're ready to merge this after the deploy to Live (to give us more time to see it on QA)

PR #348, opened 2023-04-05
Implement a FileItem parser

hartsick:
Currently we're causing an error when a user submits a file URL to Pender, either via the Check web interface or API. This means that partners can't submit links pointing to files.

While we want to eventually detect filetype using a HEAD request to the URL before fetching the file itself, this is a quick and easy way for us to do a best guess at most file types we would expect using the current flow - making a get request, then passing it off to a parser that matches the URL structure.

In order to prevent erroring on the Check API side, I also had to prevent the oembed from returning the file data. This invovled what I hope is a temporary carveout for file parsers, and telling them to use the default Oembed.

The result of all of this is that users can submit a link to a file and just see a URL with no additional metadata or oembed in the Check view, kind of like our normal fallback behavior.

CV2-2724

### DIFF
--- a/app/models/concerns/media_oembed.rb
+++ b/app/models/concerns/media_oembed.rb
@@ -34,10 +34,12 @@ module MediaOembed
     url = original_url || self.url
     if Media.valid_raw_oembed?(self.data)
       self.data['oembed'] = self.data['raw']['oembed'].merge(width: maxwidth, height: maxheight, html: Media.default_oembed_html(url, maxwidth, maxheight))
+    elsif self.provider == 'file'
+      self.data['oembed'] = Media.default_oembed(self.data, url, maxwidth, maxheight)
     else
       self.as_json if self.data.empty?
       %w(type provider).each { |key| self.data[key] = self.send(key.to_sym) }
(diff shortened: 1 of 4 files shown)


app/models/concerns/media_oembed.rb
```diff
@@ -34,10 +34,12 @@ def get_oembed_data(original_url = nil, maxwidth = nil, maxheight= nil)
     url = original_url || self.url
     if Media.valid_raw_oembed?(self.data)
       self.data['oembed'] = self.data['raw']['oembed'].merge(width: maxwidth, height: maxheight, html: Media.default_oembed_html(url, maxwidth, maxheight))
+    elsif self.provider == 'file'
```
caiosba, line 37: Can this be one of the first `if`'s so we don't process the other cases? I'm just concerned that the other cases can require more HTML parsing, which doesn't make sense for files.

hartsick: sure!

hartsick:
Current test failure is also failing on develop - think the Facebook URL is no longer returning an Oembed in the format we expected previously. Will look into independently tomorrow.

PR #350, opened 2023-04-19
[WIP] Cv2 2331 use jsonld data when parsing tiktok profiles

hartsick:


vasconsaurus:
@hartsick since you opened the PR I can't assign you as a reviewer 😅

hartsick:
@vasconsaurus hah, had never run into that before! reviewing now

test/models/parser/youtube_item_test.rb
```diff
@@ -15,8 +15,8 @@ class YoutubeItemIntegrationTest < ActiveSupport::TestCase
   end
 
   test "should store oembed data of a youtube item" do
-    RequestHelper.stubs(:get_html).returns(Nokogiri::HTML("<link rel='alternate' type='application/json+oembed' href='http://www.youtube.com/oembed?format=json&amp;url=https%3A%2F%2Fwww.youtube.com%2Fwatch%3Fv%3DmtLxD7r4BZQ' title='RubyConf Portugal 2016 - Best Of'>"))
-    m = create_media url: 'https://www.youtube.com/watch?v=mtLxD7r4BZQ'
+    RequestHelper.stubs(:get_html).returns(Nokogiri::HTML("<link rel='alternate' type='application/json+oembed' href='http://www.youtube.com/oembed?format=json&amp;url=https%3A%2F%2Fwww.youtube.com%2Fwatch%3Fv%3DfHatsiQvDWc' title='RubyConf Portugal 2016 - Best Of'>"))
```
vasconsaurus, line 18: @hartsick one thing to add here: this link `http://www.youtube.com/oembed?format=json&amp;url=https%3A%2F%2Fwww.youtube.com%2Fwatch%3Fv%3DfHatsiQvDWc` goes nowhere, but if we change it to `https://www.youtube.com/oembed?format=json&url=https%3A%2F%2Fwww.youtube.com%2Fwatch%3Fv%3DfHatsiQvDWc` it works properly.

The change is `&amp;` to `&`

For the test the one that goes nowhere worked though 🤔

test/models/parser/tiktok_profile_test.rb
```diff
@@ -44,17 +44,23 @@ def doc
 
     assert_equal '@fakeaccount', data['external_id']
     assert_equal '@fakeaccount', data['username']
-    assert_match /Huxley the Panda Puppy🐼🐶\(pandaloon \(@huxleythepandapuppy\) on TikTok/, data['description']
+    assert_match '@fakeaccount', data['title']
```
hartsick, line 47: I think we should remove these lines that were added in this test. In general we try to keep our integration tests fairly minimal because there's no guarantee that they'll fetch the HTML and parse successfully. As a result, we try to only test what will be consistent across failure and success -- in this case, it's the things we always pull from the URL (not title, description, or author name, which we try to pull from the HTML)

Instead, we try to test all of the values (fallback and also successful parse) in the unit tests below.

hartsick (commented):
Great job, @vasconsaurus! I requested a couple small changes in the comments. Once you're done with those, you should also rebase off of develop to fix the merge conflicts. When it's time to merge, we can squash all of these commits into one via Github or commandline with an interactive rebase—let me know if this is new to you and you'd like to walk through it!).

test/models/parser/tiktok_profile_test.rb
hartsick: Also, I can't request changes because it's my PR. Learning about Github permissions 😅

test/models/parser/youtube_item_test.rb
hartsick: I'm not sure what to make of that last point, but making sure the URL uses the proper characters (& instead of &amp;) makes sense to me!

app/models/parser/base.rb
```diff
@@ -69,7 +69,7 @@ def oembed_url(doc)
     attr_reader :unavailable_page
 
     # Implemented by subclasses
-    def parse_data_for_parser(doc, original_url, jsonld)
+    def parse_data_for_parser(doc, original_url, jsonld_array)
```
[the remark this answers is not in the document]
vasconsaurus: Not at all, but I do have a question. This is what that part of the code looks like:

```
    def parse_data_for_parser(doc, original_url, jsonld_array)
      raise NotImplementedError.new("Parser subclasses must implement parse_data_for_parser")
    end
```

Should the `_` be added only to `jsonld_array`, should we add to all, should we use` *` ?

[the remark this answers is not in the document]
hartsick: Oh, I'm glad you asked; I didn't realize this alert was on the base parser -- as a result, this alert is kind of irrelevant since we don't use any of those variable names in this method. Instead, it's meant to define an interface for subclasses to implement.

Because of that, I think we should leave the variable names as-is. In this case the interface that we're defining (three variables with specific meanings) is important, so we wouldn't want to use the * since it hides what we want the subclasses to pass in.

Sorry for any confusion!

vasconsaurus:
@hartsick I tried rebasing, but I think it would be helpful to to walk through it when you have the time 🙂

PR #354, opened 2023-05-12
Update production.rb to actually be used in production

hartsick:
Previously we were pulling this value from SSM, which sometimes was confusing. Now we will be reading from the committed production.rb in deployed environments.

There was nothing specific to the environment configured in the file—just what was already committed, or an unintentional variation of it—, so we don't need to add any new SSM variables.

CV2-3054

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -49,7 +49,7 @@ Rails.application.configure do
   config.log_level = :debug
 
   # Prepend all log lines with the following tags.
-  # config.log_tags = [ :subdomain, :uuid ]
+  config.log_tags = [ :subdomain, :uuid, :request_id ]
 
   # Use a different logger for distributed setups.
   # config.logger = ActiveSupport::TaggedLogging.new(SyslogLogger.new)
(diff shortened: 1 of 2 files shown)


config/environments/production.rb
```diff
@@ -49,7 +49,7 @@
   config.log_level = :debug
 
   # Prepend all log lines with the following tags.
-  # config.log_tags = [ :subdomain, :uuid ]
+  config.log_tags = [ :subdomain, :uuid, :request_id ]
```
hartsick, line 52: Updated this to match what we have on Check API, which I think is useful

production/bin/create_configs.sh
```diff
@@ -64,24 +64,6 @@ echo "Starting application configuration. Processing ENV settings."
     fi
   fi
 
-  # Populate production environment config from SSM:
```
hartsick, line 67: We no longer want to do this since it's what pulled the file out of SSM and decoded it from Base64. So, delete!

caiosba (approved):

PR #355, opened 2023-05-12
CV2-3119: protection weakened or disabled warning

vasconsaurus:
We want to remove the CodeQL warning about CSRF.

From the docs:
APIs may want to disable this behavior since they are typically designed to be state-less: that is, the request API client handles the session instead of Rails. One way to achieve this is to use the :null_session strategy instead, which allows unverified requests to be handled, but with an empty session

source:
https://api.rubyonrails.org/classes/ActionController/RequestForgeryProtection.html#method-i-verify_authenticity_token

As Caio requested, I ran an integration test, just to double check it didn't break anything, and that passed 🎉

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,5 +1,5 @@
 class ApplicationController < ActionController::Base
-  protect_from_forgery with: :null_session
+  skip_before_action :verify_authenticity_token
   
   def process_action(*)
     if PenderConfig.get('memory_report', false).to_s == 'true'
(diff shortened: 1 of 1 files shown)


caiosba (approved):

caiosba (dismissed):
@vasconsaurus , looks like that change was not enough (and we also need to remove the line `  skip_before_action :verify_authenticity_token`... let's try something else.

vasconsaurus:
Talked to @caiosba and we decided to re-open this to remove `protect_from_forgery`. 

When we add `protect_from_forgery` and then `skip_before_action`, we make the first one pointless, and it can cause confusion since it doesn't have a real reason to be there. Since this is a stateless API we don't need to use `protect_from_forgery`, so we can only keep `skip_before_action`.

hartsick (approved):
looks good to me! agree this is a lot less confusing

PR #364, opened 2023-07-11
3019 - add safe operator to avoid NoMethodError (kwai)

vasconsaurus:
## Description

This was surfaced through audit hour.

Inside the Kwai parser, we added a safe operator before `strip` in `get_kwai_text_from_tag`. We did so because that might be `nil` and we don't want it to return an error.

References: cv2-3019

## How has this been tested?

I wasn't really able to fully test it. But this also seems pretty straight forward. Anyway, on the tests: 
- I wanted to test it against the url that caused the error on sentry, but the sentry issue is no longer there. I'm not sure why 🤔 I'm not sure how to reproduce this.
- We were skipping the kwai integration test, because it was broken. I ran it anyway and it seems to be working now. @hartsick should we go back to running this test?

### DIFF
--- a/app/models/parser/kwai_item.rb
+++ b/app/models/parser/kwai_item.rb
@@ -30,7 +30,7 @@ module Parser
     end
   
     def get_kwai_text_from_tag(doc, selector)
-      doc&.at_css(selector)&.text&.to_s.strip
+      doc&.at_css(selector)&.text&.to_s&.strip
     end
   end
 end
--- a/test/models/parser/kwai_test.rb
+++ b/test/models/parser/kwai_test.rb
@@ -2,8 +2,6 @@ require 'test_helper'
 
 class KwaiIntegrationTest < ActiveSupport::TestCase
   test "should parse Kwai URL" do
-    skip "Kwai parsing is not currently working; pending until we look into"
-
     m = create_media url: 'https://s.kw.ai/p/1mCb9SSh'
     data = m.as_json
     assert_equal 'Reginaldo Silva2871', data['username']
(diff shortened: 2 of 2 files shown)


vasconsaurus:
@hartsick before I merge this, I just wanted to make sure: should we go back to running the integration test or should we keep the skip?

hartsick:
If it's passing, let's run it!

vasconsaurus:
@hartsick it failed but the test that didn't pass wasn't the one that was being skipped before.
we were skipping the integration test, and one of the unit tests failed (`rails test test/models/parser/kwai_test.rb:49`) 🤔

vasconsaurus:
@hartsick when I removed the safe operator that test passed locally, but I don't know why it passed before and stopped now. So I'm not quite sure what is best to do here

hartsick:
Hmmm thanks for bringing these up. I think we may want to actually change this approach a bit in response.

I think this is another case where it might be useful to monitor this case to understand if the expectations we have against Kwai format have changed in Sentry. But it isn't a meaningful error if doc is nil. So, here is my suggestion:

Return from the parsing function early if doc is nil, since we can't get any meaningful data out of it and it doesn't tell us if our parsing attempt is broken (just fetching attempt). Then, in get_kwai_text_from_tag keep it as-is, and continue raising the error. We'll need to update the test to reflect the different error condition, and write a new test to cover the additional nil return case. Should be a good opportunity for test driving this change!

vasconsaurus:
@hartsick I tried returning if a `doc` is `nil`, but I ran into a few things:

- the `doc` seems to never be `nil`, even if the page doesn't exist it returns a nokogiri object, like this: `#<Nokogiri::HTML4::Document:0x93d0 name="document" children=[#<Nokogiri::XML::DTD:0x93e4 name="html">, #<Nokogiri::XML::Element:0x93f8 name="html" children=[#<Nokogiri::XML::Element:0x940c name="body" children=[#<Nokogiri::XML::Element:0x9420 name="p" children=[#<Nokogiri::XML::Text:0x9754 "link length is not valid">]>]>]>]>`
- I could check if the `selector` is `nil`, but then it would be kinda the same as adding the `safe operator`. We wouldn't get to `get_kwai_text_from_tag` and our test would fail.
- Caio was talking about how pender should always return something, so I was wondering if we should return early?
- I also was able to find the issue in Sentry again, and this is a very specific link (seems like a link to send money?credits to a user). This happened once in 3 months, and when running the link I didn't get the error. So I was thinking if we should change this at all?

what do you think?

hartsick:
@vasconsaurus re #1 and #2, noted!

For #3, Pender will return something even if we return early from the parser, since the setting of fallbacks (title and any other required information) happens in [media.rb](https://github.com/meedan/pender/blob/develop/app/models/media.rb#L77) after we try parsing.

For #4, I agree it's not high-priority, but worth fixing if we're already here and the solution is straightforward (defer to you on that).

What would you like to do?

vasconsaurus:
> For https://github.com/meedan/pender/pull/3, Pender will return something even if we return early from the parser, since the setting of fallbacks (title and any other required information) happens in [media.rb](https://github.com/meedan/pender/blob/develop/app/models/media.rb#L77) after we try parsing.

derp 🤦‍♀️ 

So I think it makes sense to return if there is no selector, and update our test. Which is what you were saying before 😅

hartsick:
haha no worries!! I had to go back and check it myself to make sure we didn't just do that when an exception was raised.

hartsick (approved):
🎉

PR #367, opened 2023-07-17
NoMethodError in tiktok parser (sentry / audit hour)

vasconsaurus:
## Description

When looking into this issue on Sentry, we found that we had two issues here. On some events the page didn't exist, on others it wasn't either a profile nor an item, but a tags result page.

**For the page that does not exist:** adding a safe operator inside the parser seems to be enough. Now instead of getting the `NoMethodError` we get `URL Not Found`, which I think is a better error here.

**For the tag page:** I added some logic to deal with that in `tiktok_item`, and added an @ to the regex inside `profile`, to make sure this goes through the item parser.

References: cv2-3124

## How has this been tested?

Manually and through tiktok's tests. I also updated the [test](https://github.com/meedan/pender/blob/fix-3124-tiktok_parser-undefined_method/test/models/parser/tiktok_item_test.rb#L53) ("matches known URL patterns, and returns instance on success") to test for the new tag pattern.


app/models/parser/tiktok_item.rb
```diff
@@ -22,10 +23,21 @@ def parse_data_for_parser(doc, _original_url, _jsonld_array)
 
       handle_exceptions(StandardError) do
         @parsed_data[:raw][:api] = @parsed_data[:raw][:oembed] = get_tiktok_api_data(url)
-        match = url.match(TIKTOK_ITEM_URL)
+        
+        if url.match(TIKTOK_ITEM_URL)
+          match = url.match(TIKTOK_ITEM_URL)
+          username = match['username']
+          external_id = match['id']
+        end
```
hartsick, line 31: suggestion: I think an elsif might be a little clearer here, if we expect it to be one or the other

hartsick (approved):
Looks good! I left one small comment just for clarity.

I also have one small question - is there any unique or additional parsing we need for the tags page to surface relevant information for tags (vs items), or is the parsing work we already do for Tiktok items enough?

vasconsaurus:
We could return the `tag` as `tag`, instead of only in title, or maybe a `boolean` saying it's a tag page? `Username`, `author` and `id` are empty, since there are none. But I don't think there is a lot of useful information in the tags results page 🤔  . What do you think? It returns something like this:

```
   "api"=>
    {"version"=>"1.0",
     "type"=>"rich",
     "title"=>"#ball",
     "author_url"=>"",
     "author_name"=>"",
     "width"=>"100%",
     "height"=>"100%",
     "html"=>
      "<blockquote class=\"tiktok-embed\" cite=\"https://www.tiktok.com/tag/ball\" data-tag-id=\"ball\" data-embed-from=\"oembed\" data-embed-type=\"tag\" style=\"max-width: 780px; min-width: 288px;\" > <section> <a target=\"_blank\" href=\"https://www.tiktok.com/tag/ball?refer=hashtag_embed\">#ball</a> </section> </blockquote> <script async src=\"https://www.tiktok.com/embed.js\"></script>",
     "provider_url"=>"https://www.tiktok.com",
     "provider_name"=>"TikTok",
     "embed_product_id"=>"ball",
     "embed_type"=>"hashtag"}},
 "description"=>"#ball",
 "username"=>"",
 "external_id"=>"",
 "title"=>"#ball",
 "picture"=>nil,
 "author_url"=>"",
 "html"=>
  "<blockquote class=\"tiktok-embed\" cite=\"https://www.tiktok.com/tag/ball\" data-tag-id=\"ball\" data-embed-from=\"oembed\" data-embed-type=\"tag\" style=\"max-width: 780px; min-width: 288px;\" > <section> <a target=\"_blank\" href=\"https://www.tiktok.com/tag/ball?refer=hashtag_embed\">#ball</a> </section> </blockquote> <script async src=\"https://www.tiktok.com/embed.js\"></script>",
 "author_name"=>"",
 "tag"=>"ball"}
```

app/models/parser/tiktok_item.rb
```diff
@@ -22,16 +23,27 @@ def parse_data_for_parser(doc, _original_url, _jsonld_array)
...
+          username = ''
+          external_id = ''
+          tag = url.match(TIKTOK_TAG_URL)['tag']
+        end
+
         @parsed_data.merge!({
-          username: match['username'],
-          external_id: match['id'],
+          username: username,
+          external_id: external_id,
```
melsawy, line 39: `username` , `external_id` or `tag`  will be undefined in case there is no match for `TIKTOK_ITEM_URL` or `TIKTOK_ITEM_URL` so I suggest to define `username`, `external_id ` and `tag` with empty strings outside the if..else block (i.e. before line 26) and fill the right values inside `if ... elseif ... end` block

melsawy (changes requested):

app/models/parser/tiktok_item.rb
vasconsaurus: I added an 'else' in case there is no match and tested with an url. What do you think?

hartsick:
@vasconsaurus 
> We could return the `tag` as `tag`, instead of only in title, or maybe a `boolean` saying it's a tag page? `Username`, `author` and `id` are empty, since there are none. But I don't think there is a lot of useful information in the tags results page 🤔 . What do you think? It returns something like this:
> 
> ```
>    "api"=>
>     {"version"=>"1.0",
>      "type"=>"rich",
>      "title"=>"#ball",
>      "author_url"=>"",
>      "author_name"=>"",
>      "width"=>"100%",
>      "height"=>"100%",
>      "html"=>
>       "<blockquote class=\"tiktok-embed\" cite=\"https://www.tiktok.com/tag/ball\" data-tag-id=\"ball\" data-embed-from=\"oembed\" data-embed-type=\"tag\" style=\"max-width: 780px; min-width: 288px;\" > <section> <a target=\"_blank\" href=\"https://www.tiktok.com/tag/ball?refer=hashtag_embed\">#ball</a> </section> </blockquote> <script async src=\"https://www.tiktok.com/embed.js\"></script>",
>      "provider_url"=>"https://www.tiktok.com",
>      "provider_name"=>"TikTok",
>      "embed_product_id"=>"ball",
>      "embed_type"=>"hashtag"}},
>  "description"=>"#ball",
>  "username"=>"",
>  "external_id"=>"",
>  "title"=>"#ball",
>  "picture"=>nil,
>  "author_url"=>"",
>  "html"=>
>   "<blockquote class=\"tiktok-embed\" cite=\"https://www.tiktok.com/tag/ball\" data-tag-id=\"ball\" data-embed-from=\"oembed\" data-embed-type=\"tag\" style=\"max-width: 780px; min-width: 288px;\" > <section> <a target=\"_blank\" href=\"https://www.tiktok.com/tag/ball?refer=hashtag_embed\">#ball</a> </section> </blockquote> <script async src=\"https://www.tiktok.com/embed.js\"></script>",
>  "author_name"=>"",
>  "tag"=>"ball"}
> ```

I think that we shouldn't return tag as a new top-level key from the API, since it would be specific to TikTok items and the frontend won't do anything with it by default. our top-level keys should be consistent for all parsers in the API response.

I think we can work in the keys we have, though! right now the title is a bit confusing since it's just #ball, so my suggestion is if we see that we've gotten a tag page let's modify the title to be `Tag <original-title>` (in this case it would be `Tag #ball`) or something similar. what do you think?

test/models/parser/tiktok_item_test.rb
```diff
@@ -56,6 +56,9 @@
     
     match_one = Parser::TiktokItem.match?('https://www.tiktok.com/@fakeaccount/video/abcdef?a=1')
     assert_equal true, match_one.is_a?(Parser::TiktokItem)
+
+    match_one = Parser::TiktokItem.match?('https://www.tiktok.com/tag/randomtag')
```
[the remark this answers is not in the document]
hartsick: this is a false positive that codeql gives us because it doesn't understand our .match? is different than regex .match?

vasconsaurus:
@hartsick @melsawy please take a look this one when you have the time ☺️

app/models/parser/tiktok_item.rb
```diff
@@ -22,12 +23,26 @@ def parse_data_for_parser(doc, _original_url, _jsonld_array)
...
-        match = url.match(TIKTOK_ITEM_URL)
+        if url.match(TIKTOK_ITEM_URL)
+          match = url.match(TIKTOK_ITEM_URL)
+          username = match['username']
+          external_id = match['id']
+          title = parsed_data['raw']['api']['title']
+        elsif url.match(TIKTOK_TAG_URL)  
+          username = ''
+          external_id = ''
+          title = "Tag: " + url.match(TIKTOK_TAG_URL)['tag']
```
hartsick, line 34: could you add a test here? seems like good behavior to capture. think we already catch the other cases through our existing tests

hartsick (approved):
I think we should add one test to reflect the new title setting behavior. otherwise looks good!

app/models/parser/tiktok_item.rb
vasconsaurus: @hartsick I tried adding two tests one to test if the 'Tag: ' is added, and another to test if `external_id` and `username` are empty strings for unmatched URL pattern.

The one for the tag seems to work, but the other one doesn't.

At first I tried writing something like this
```
test "sets 'external_id' and 'username' as empty string for unmatched URL pattern" do
    WebMock.stub_request(:get, 'https://www.tiktok.com/search?q=abcdef').to_return(status: 200, body: response_fixture_from_file('tiktok-item-page.html'))
    
    data = Parser::TiktokItem.new('https://www.tiktok.com/search?q=abcdef').parse_data(doc)

    assert_equal '', data['title']
    assert_equal '', data['username']
end
```

I got an error that told me to use the `oembed`. I changed it, and it passed.
But it failed on Travis, so I ran it again locally, and it's currently failing as well.
I don't know why it passed the first couple tries.

What is the correct way to write this? Should we be testing for this?

hartsick (approved):
faaaaantastic 🎉

PR #371, opened 2023-07-27
3309 - Update pender to use twitter's v2 API

vasconsaurus:
## Description

We were using Twitter's v1.1 API, which has since June stopped working, so we are updating to use v2. 

**This work included**
1. creating a concern to deal with the requests to the twitter api (`provider_twitter`)
2. adding error handling and sending notifications to sentry
3. creating, updating and removing tests
4. updating `twitter_item` and `twitter_profile` parsers 
5. updating our config (including the example.config)
6. removing the twitter gem
7. removing twitter urls from tests that didn't need it to be from twitter

References: cv2-3309

## How has this been tested?

1. I made some curl requests using the api
```
curl --request GET 'https://api.twitter.com/2/users/by?usernames=NASAWebb&user.fields=profile_image_url,name,username,description,created_at' --header "Authorization: Bearer $TWITTER_BEARER"
```
```
curl --request GET 'https://api.twitter.com/2/tweets?ids=1685816399292223491&tweet.fields=author_id,created_at,text&expansions=author_id,attachments.media_keys&user.fields=profile_image_url,username,url&media.fields=url' --header "Authorization: Bearer $TWITTER_BEARER"
```
2. Added unit and integration tests to both `twitter_profile` and `twitter_item`

The tests. They had to be re-written, so extra attention might be warranted. Also, please let me know if there is any scenario missing that we should be testing.

Currently Travis tests are failing because of some changes to Facebook. Those do not affect this PR, and we have a ticket to deal with them separately.

## Checklist

- [x] I have performed a self-review of my own code
- [x] I have added unit and feature tests, if the PR implements a new feature or otherwise would benefit from additional testing
- [x] I have added logging, exception reporting, and custom tracing with any additional information required for debugging
- [x] My changes generate no new warnings


vasconsaurus:
@caiosba @hartsick @melsawy  

I started working on updating the api, there were quite a few changes already, so I think It might be a good idea for us to talk through a few things:
- I started by making the new `Twitter Client` a class. But I noticed that we have something similar as a concern for instagram ([instagram_provider](https://github.com/meedan/pender/blob/0551496671198295ad0ec3c06fe0d8999cd47b21/app/models/concerns/provider_instagram.rb)), and we have a [twitter_provider](https://github.com/meedan/pender/blob/73078c3b6f41da5ceb826da17c573ce4f6119b27/app/models/concerns/provider_twitter.rb) . So I added the methods on the new `Twitter Client` class to the `twitter_provider`, and removed the class. WDYT? Should this be a class, a concern?
- I tried to add some error handling [here](https://github.com/meedan/pender/blob/3309-update-twitter-v2/app/models/concerns/provider_twitter.rb#L61), but I'm not too sure if this is correct/enough
- note: now when the api returns the `user url`, it returns the `url` the user has in their profile, which is an `external url`, not the twitter one (we can get the twitter profile url other ways, but I thought this was important to note)

caiosba:
Great work so far @vasconsaurus !

> @caiosba @hartsick @melsawy
> 
> I started working on updating the api, there were quite a few changes already, so I think It might be a good idea for us to talk through a few things:
> 
> * I started by making the new `Twitter Client` a class. But I noticed that we have something similar as a concern for instagram ([instagram_provider](https://github.com/meedan/pender/blob/0551496671198295ad0ec3c06fe0d8999cd47b21/app/models/concerns/provider_instagram.rb)), and we have a [twitter_provider](https://github.com/meedan/pender/blob/73078c3b6f41da5ceb826da17c573ce4f6119b27/app/models/concerns/provider_twitter.rb) . So I added the methods on the new `Twitter Client` class to the `twitter_provider`, and removed the class. WDYT? Should this be a class, a concern?

I don't have a strong opinion. I think that a separate class makes sense in terms of encapsulation, it's easier to be tested, etc. But at the end of the day, it's going to be used only by `TwitterItemParser` and `TwitterProfileParser`, and they both include the `TwitterProvider` concern, so it's a valid place for that code too. If it gets too big, then I think it would make more sense to move it to a class.

> * I tried to add some error handling [here](https://github.com/meedan/pender/blob/3309-update-twitter-v2/app/models/concerns/provider_twitter.rb#L61), but I'm not too sure if this is correct/enough

It's looking good! It's basically two cases really: any exception or any response code that is not successful. So, I would consider successful `response.code.to_i < 400`.  I have a question there: is `response.message` valid? I just knew about `response.body` for a HTTP response. I also think it's important to include the `response.code` in that exception. Finally, I think that for now it's better to `rescue StandardError`, since there might be cases that we don't know yet... we can refine further as we see such cases happening in Sentry.

> * note: now when the api returns the `user url`, it returns the `url` the user has in their profile, which is an `external url`, not the twitter one (we can get the twitter profile url other ways, but I thought this was important to note)

For a Twitter profile URL, we can build this URL as `https://twitter.com/ + Twitter handle`, and that other URL we can store in the payload anyway.

caiosba:
> @caiosba Regarding your comments on error handling, do you think this makes sense [9343c39](https://github.com/meedan/pender/commit/9343c392567c6975ab2e2bf427050395779155f3) ?
> 
> > is response.message valid?
> 
> yes slightly_smiling_face For example, if we send an invalid token the `response.message` is `Unauthorized`

Ah, OK! Thanks! Good to know :)

Yes, that commit makes sense, but given what you just explained about `response.message`, I think it's also useful to include `response.body`... for example, a `response.message` can be `Unauthorized` but the `response.body` will probably include more details about the reason, for example, "token expired" or "insufficient permissions", etc.

vasconsaurus:
@caiosba @hartsick @melsawy I have some questions about some of our twitter_item tests 😬 

**On tests on error handling [here](https://github.com/meedan/pender/blob/aa8f72731f5dbf17e02636bd5a21169ac4167f5c/test/models/parser/twitter_item_test.rb#L172C3-L172C3) and [here](https://github.com/meedan/pender/blob/aa8f72731f5dbf17e02636bd5a21169ac4167f5c/test/models/parser/twitter_item_test.rb#L196)**
Caio suggested we just use StandardError for now and add more specific errors after tracking on sentry. This however changes how we were tracking before, also we were using the twitter gem, which we won't be using anymore. New proposed error handling behavior [here](https://github.com/meedan/pender/blob/2054e9a44ddbc8c06779cfdcbc8dfa2bde3ef6c3/app/models/concerns/provider_twitter.rb#L60)

- Should we add functionality to make these tests pass, if so how could we do that? 
- Or should we change those tests, and, if so, how could we do that? What do we want to test?

**On setting the author url, [here](https://github.com/meedan/pender/blob/aa8f72731f5dbf17e02636bd5a21169ac4167f5c/test/models/parser/twitter_item_test.rb#L218)**
We have this test that checks if when the user information can't be retrieved we set the author_url to twitter.com. But, since this is an tweet, we get the user from the url and get the author_url [here](https://github.com/meedan/pender/blob/2054e9a44ddbc8c06779cfdcbc8dfa2bde3ef6c3/app/models/parser/twitter_item.rb#L66). So I'm wondering if this test makes sense, if so, what would cause this to happen?

caiosba:
@vasconsaurus , my answers inline below:

> * Should we add functionality to make these tests pass, if so how could we do that?
> * Or should we change those tests, and, if so, how could we do that? What do we want to test?

I don't think we need to mimic the same tests, since the error handling is different. My suggestion is to only test that when the HTTP response is not successful (so, in the range of 400 or 500, for example), we notify Sentry and raise our custom exception class.

> **On setting the author url, [here](https://github.com/meedan/pender/blob/aa8f72731f5dbf17e02636bd5a21169ac4167f5c/test/models/parser/twitter_item_test.rb#L218)** We have this test that checks if when the user information can't be retrieved we set the author_url to twitter.com. But, since this is an tweet, we get the user from the url and get the author_url [here](https://github.com/meedan/pender/blob/2054e9a44ddbc8c06779cfdcbc8dfa2bde3ef6c3/app/models/parser/twitter_item.rb#L66). So I'm wondering if this test makes sense, if so, what would cause this to happen?

I don't think it makes sense that way it was before on `develop`... I agree with the comment here, we can just set the `author_url` to be `https://twitter.com/<user_handle>`.

.travis.yml
```diff
@@ -1,6 +1,6 @@
 language: minimal
 before_install:
-- openssl aes-256-cbc -K $encrypted_8cec9149bf7a_key -iv $encrypted_8cec9149bf7a_iv -in config/config.yml.enc -out config/config.yml -d
+-  openssl aes-256-cbc -K $encrypted_3491736328a2_key -iv $encrypted_3491736328a2_iv -in config/config.yml.enc -out config/config.yml -d
```
caiosba, line 3: Minor change suggestion: We can delete this leading space

app/models/parser/base.rb
```diff
@@ -166,13 +157,7 @@ def get_twitter_metadata
...
     def twitter_author_url(username)
       return if bad_username?(username)
-      begin
-        twitter_client.user(username)&.url&.to_s
-      rescue Twitter::Error => e
-        PenderSentry.notify(e, url: url, username: username)
-        Rails.logger.warn level: 'WARN', message: "[Parser] #{e.message}", username: username, error_class: e.class
-        nil
-      end
+        "https://twitter.com/" + username[1..]
```
caiosba, line 160: Minor change suggestion: Remove the leading space
Question: `username[1..]` is because the first character is always `@`? If so, can we please add a comment just to make it super clear?

app/models/parser/twitter_item.rb
```diff
@@ -19,47 +19,62 @@ def patterns
     # Main function for class
     def parse_data_for_parser(_doc, _original_url, _jsonld_array)
       @url.gsub!(/(%23|#)!\//, '')
+      @url.gsub!(/\s/, '')
       @url = replace_subdomain_pattern(url)
       parts = url.match(TWITTER_ITEM_URL)
+      
       user, id = parts['user'], parts['id']
       
       @parsed_data['raw']['api'] = {}
```
caiosba, line 28: Minor change suggestion: I think this initialization is not needed.

app/models/parser/twitter_profile.rb
```diff
@@ -19,25 +19,37 @@ def patterns
...
+      @url = replace_subdomain_pattern(url)      
+      username = compare_patterns(@url, self.patterns, 'username')
 
-      @parsed_data[:raw][:api] = {}
-      handle_twitter_exceptions do
-        @parsed_data[:raw][:api] = twitter_client.user(username).as_json
-        picture_url = parsed_data[:raw][:api][:profile_image_url_https].gsub('_normal', '')
-        set_data_field('picture', picture_url)
-        set_data_field('author_picture', picture_url)
+      @parsed_data[:raw][:api] = {}
```
caiosba, line 26: Minor change suggestion: I think this initialization is not needed

spec/requests/api/medias_spec.rb
```diff
@@ -163,48 +163,48 @@
...
+      #     expect(data['message']).to eq(123)
+      #   end
+
+      #   after do
+      #     allow_any_instance_of(Twitter::REST::Client).to receive(:user).and_call_original
+      #     allow_any_instance_of(Twitter::Error::TooManyRequests).to receive(:rate_limit).and_call_original
+      #   end
+
+      #   include_context 'generate examples'
+      # end
```
caiosba, line 207: Question: Why these tests are commented? If we don't need them anymore, let's just delete these lines.

test/integration/parsers/twitter_item_test.rb
```diff
@@ -0,0 +1,15 @@
+require 'test_helper'
+
+class TwitterItemIntegrationTest < ActiveSupport::TestCase
+  test "should parse tweet" do
+    # skip("twitter api key is not currently working")
```
caiosba, line 5: Suggestion: We can delete this line

test/models/parser/page_item_test.rb
```diff
@@ -107,7 +106,7 @@ class PageItemIntegrationTest < ActiveSupport::TestCase
   end
 
   test "should parse urls without utf encoding" do
-    urls = [
+       urls = [
```
caiosba, line 109: Minor suggestion: Delete the leading spaces here

test/models/parser/page_item_test.rb
```diff
@@ -129,6 +128,10 @@ class PageItemIntegrationTest < ActiveSupport::TestCase
   end
 
   test "should handle error when cannot get twitter url" do
```
caiosba, line 130: I think we can delete this test.

test/models/parser/twitter_profile_test.rb
```diff
@@ -71,62 +60,125 @@ def empty_doc
...
+    stub_profile_lookup.returns(twitter_profile_response_success)
+    
+    data = Parser::TwitterProfile.new('https://www.twitter.com/fake_user').parse_data(empty_doc)
+    
+    assert_equal 'fake_user', data['external_id']
+    assert_equal '@fake_user', data['username']
+    assert_match /The world's most powerful space telescope/, data['description']
+    assert_match 'fake_user', data['title']
+    assert_match 'Fake User', data['author_name']
+
```
caiosba, line 121: Minor suggestion: please delete these whitespaces (lines 113, 115 and 121)

test/models/parser/twitter_profile_test.rb
```diff
@@ -71,62 +60,125 @@ def empty_doc
...
+    assert_match /The world's most powerful space telescope/, data['description']
+    assert_match 'fake_user', data['title']
+    assert_match 'Fake User', data['author_name']
+    
+    assert_match 'https://twitter.com/fake_user', data['url']
+    assert_match /pbs.twimg.com\/profile_images\/685182791496134658\/Wmyak8D6.jpg/, data['picture']
+    assert_match /pbs.twimg.com\/profile_images\/685182791496134658\/Wmyak8D6.jpg/, data['author_picture']
     assert_not_nil data['published_at']
-
+
```
caiosba, line 126: Minor suggestion: please delete theis whitespace

caiosba (changes requested):
It's looking really good, amazing work @vasconsaurus ! I left some questions and suggestions. Other than these, I also suggest:

- Please add integration tests for the cases of failure too (for example, a tweet ID that doesn't exist and a profile that doesn't exist)
- Please add `twitter_bearer_token` to `config.yml.example`

app/models/parser/twitter_item.rb
```diff
@@ -19,47 +19,62 @@ def patterns
...
-      handle_twitter_exceptions do
-        @parsed_data['raw']['api'] = twitter_client.status(id, tweet_mode: 'extended').as_json
+      @parsed_data['raw']['api'] = tweet_lookup(id)
+
+      @parsed_data[:error] = parsed_data['raw']['api']['errors']
+
+      if @parsed_data[:error] 
+        title = ''
```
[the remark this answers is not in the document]
vasconsaurus: This was my mistake, wil update it to be the url when there is an error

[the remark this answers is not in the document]
hartsick: if title is empty in a parser class, we set title to the URL in media after the parse_data method returns. so this is how I'd expect it to be written! no need to change.

app/models/concerns/provider_twitter.rb
```diff
@@ -3,22 +3,57 @@
...
-      yield
-    rescue Twitter::Error::TooManyRequests => e
-      raise Pender::Exception::ApiLimitReached.new(e.rate_limit.reset_in)
-    rescue Twitter::Error => error
-      PenderSentry.notify(error, url: url)
-      @parsed_data[:raw][:api] = { error: { message: "#{error.class}: #{error.code} #{error.message}", code: Lapis::ErrorCodes::const_get('INVALID_VALUE') }}
-      Rails.logger.warn level: 'WARN', message: "[Parser] #{error.message}", url: url, code: error.code, error_class: error.class
-      return
+      response = http.request(request)
+      raise ApiError.new("#{response.class}: #{response.code} #{response.message} - #{response.body}") unless response.code.to_i < 400
```
hartsick, line 52: **suggestion**: I think we should remove response.body from this error message construction because it should give us a unique error each time in Sentry. It would be helpful to make sure that the response body is sent as an attribute to Sentry, though, for easier debugging!

app/models/concerns/provider_twitter.rb
```diff
@@ -3,22 +3,57 @@
...
+      JSON.parse(response.body) 
+    rescue StandardError => e
+      PenderSentry.notify(e, url: url)
+      raise ApiError.new("#{e.class}: #{e.message}")
```
hartsick, line 56: **question**: do we want to just return here rather than re-raising an error? that seems to match the previous behavior.

app/models/parser/base.rb
hartsick: **suggestion** I'd actually like to make a different suggestion to make the code a bit clearer to me - I'd suggest using gsub (i.e. username.gsub("@",""). that supports both a case where username is prefixed with @ and isn't, and is more in line with what we do in other classes

app/models/parser/twitter_item.rb
```diff
@@ -19,47 +19,63 @@ def patterns
...
       
-      @parsed_data['raw']['api'] = {}
-      handle_twitter_exceptions do
-        @parsed_data['raw']['api'] = twitter_client.status(id, tweet_mode: 'extended').as_json
+      @parsed_data['raw']['api'] = tweet_lookup(id)
+
+      @parsed_data[:error] = parsed_data['raw']['api']['errors']
+
+      if @parsed_data[:error] 
+        title = url
```
hartsick, line 33: **suggestion** I think that we don't need most of this block, since by default the values would be unset and title will be set outside of this class if it's nil. if you want to set the author_url and author_name, maybe we just keep that?

app/models/parser/twitter_item.rb
```diff
@@ -19,47 +19,63 @@ def patterns
...
+        description = ''
+        author_picture = ''
+        published_at = ''
+        html = ''
+        author_name = user
+        author_url = get_author_url(user)
+      elsif @parsed_data[:error].nil?
+        raw_data = parsed_data['raw']['api']['data'][0]
```
hartsick, line 41: **suggestion**: for accessing deeply nested data where there might not always be a value, I'd suggest using dig, ie. `parsed_data.dig('raw','api','data',0)`. this will return nil if data is missing, whereas the current line and some others below would raise an error for invalid access.

app/models/parser/twitter_item.rb
```diff
@@ -19,47 +19,63 @@ def patterns
...
+        description = title
+        author_picture = raw_user_data['profile_image_url'].gsub('_normal', '')
+        published_at = raw_data['created_at']
+        html = html_for_twitter_item(url)
+        author_name = raw_user_data['name']
+        author_url = get_author_url(user) || raw_user_data['url'] || RequestHelper.top_url(url)
       end
-      @parsed_data[:error] = parsed_data.dig(:raw, :api, :error)
+ 
       @parsed_data.merge!({
```
hartsick, line 53: I might suggest just setting the values on parsed_data in the if/else block above since there is so little data we want to set in the error case. if we need to set any default values (like picture), then consider doing it before the if/else block, since the merge! would overwrite them when called.

app/models/parser/twitter_item.rb
```diff
@@ -19,47 +19,63 @@ def patterns
...
-      picture_url.gsub('_normal', '') if picture_url
+    def get_author_url(user)
+      'https://twitter.com/' + user
     end
 
-    def picture_url(data)
-      item_media = data.dig('raw', 'api', 'entities', 'media')
-      (item_media.dig(0, 'media_url_https') || item_media.dig(0, 'media_url')) if item_media
+    def get_twitter_item_picture(parsed_data)
+      return unless parsed_data['raw']['api']['includes']
```
hartsick, line 73: **suggestion** use dig here to avoid getting an error when accessing deeply nested data

app/models/parser/twitter_profile.rb
```diff
@@ -19,25 +19,38 @@ def patterns
 
     # Main function for class
     def parse_data_for_parser(doc, _original_url, _jsonld_array)
-      @url = replace_subdomain_pattern(url)
-      username = url.match(/^https?:\/\/(www\.)?twitter\.com\/([^\/]+)$/)[2]
+      @url.gsub!(/\s/, '')
```
hartsick, line 22: **question** are we seeing whitespace in the URL here? curious what this is needed for. same question for in twitter item

app/models/parser/twitter_profile.rb
```diff
@@ -19,25 +19,38 @@ def patterns
...
+
+        picture = raw_data['profile_image_url'].gsub('_normal', '')
+        author_name = raw_data['name']
+        description = raw_data['description'].squish
+        published_at = raw_data['created_at']
       end
-      @parsed_data[:error] = parsed_data.dig(:raw, :api, :error)
-      set_data_field('title', parsed_data.dig(:raw, :api, :name), username)
+ 
       @parsed_data.merge!({
```
hartsick, line 44: **suggestion** same comments for this as in twitter item above. it's just a small thing, but think it'll help with readability since it'll be less code!

spec/requests/api/medias_spec.rb
```diff
@@ -163,49 +163,6 @@
         end
       end
 
-      # response '429', 'API limit reached' do
```
hartsick, line 166: **question** can we just delete these?

test/integration/parsers/twitter_item_test.rb
hartsick, line 1: yay for breaking this out into its own file

test/models/parser/page_item_test.rb
```diff
@@ -397,9 +380,7 @@ def fake_twitter_user
   end
 
   test "sets author name as author_name, username, and then title" do
-    Twitter::REST::Client.any_instance.stubs(:user).returns(fake_twitter_user)
-
-    doc = Nokogiri::HTML(<<~HTML)
```
hartsick, line 402: small thing, but I think this should be un-indented

test/models/parser/page_item_test.rb
```diff
@@ -491,21 +470,19 @@ def fake_twitter_user
   end
 
   test "uses twitter URL from twitter metadata for author_url (and not username) if valid" do
-    Twitter::REST::Client.any_instance.stubs(:user).returns(fake_twitter_user)
-
+    # it's getting the author_url from the metadata, but because of how we are now building the url it fails that assert
```
hartsick, line 473: let's go ahead and just delete the assertion and comment, rather than commenting! but we should make sure to test the author_url construction somewhere (this may be happening already). can be its own test if that's easiest.

test/models/parser/page_item_test.rb
```diff
@@ -539,4 +516,15 @@ def fake_twitter_user
     oembed_url = Parser::PageItem.new('https://example.com').oembed_url(oembed_doc)
     assert_equal "https://example.com/oembed", oembed_url
   end
+
+  test "shouldn't break when cannot get twitter url" do
+  # I'm trying to write a substitute to "should handle error when cannot get twitter url"
```
hartsick, line 521: for this test, I think we should attempt to model what happens if the request actually errors, or returns a response code we dislike. I could see it actually being a couple of tests:
1. mocking our request to Twitter to return an error code and message that is a 422, 500, etc. basically anything that isn't a 200 or 301 (<400, which is how we have it written in the implementation). then could test that the error gets sent to sentry, and an error hash is returned with a meaningful value, and the parser returns successfully. this is similar to the previous test, but relies on us to determine when there is an error and actually returns an error message (I don't know why we didn't do that before)
2. raise an exception from the HTTP request, and test our handling. we'll need to understand what exceptions we expect to be raised, which involves looking at documentation for Net::HTTP that we're using to make the request. I know `Net::ReadTimeout` is an example, so we could force Net::HTTP to raise that exception and test our handling of it (at the least, exit gracefully)

I think there are examples in the code base of both approaches, though may be in non-parser classes. happy to talk more about them if needed!

test/models/parser/twitter_profile_test.rb
```diff
@@ -71,62 +60,123 @@ def empty_doc
...
+  test "it makes a get request to the user lookup by username endpoint and raises an error when 401 is returned" do
+    stub_configs({'twitter_bearer_token' => 'test' })
+
+    WebMock.stub_request(:get, "https://api.twitter.com/2/users/by")
+      .with(query: query)
+      .with(headers: { "Authorization": "Bearer test" })
+      .to_return(status: 401)
+
+    assert_raises ProviderTwitter::ApiError do
```
hartsick, line 85: this is a really well-written test, but I think we want this to *not* raise the error, since we expect our parsers to always return an object and handle their internal errors themselves. this is why we normally wrap parse_data internals in an `handle_exceptions(StandardError)` block. I think that could be a good approach here, too! instead, we should do something like the test below where we verify the exception is sent to Sentry

test/models/parser/twitter_profile_test.rb
```diff
@@ -71,62 +60,123 @@ def empty_doc
...
     sentry_call_count = 0
     arguments_checker = Proc.new do |e|
       sentry_call_count += 1
-      assert_equal Twitter::Error, e.class
     end
-
+    
     PenderSentry.stub(:notify, arguments_checker) do
-      data = Parser::TwitterProfile.new('https://www.twitter.com/fakeaccount').parse_data(nil)
+      assert_raises ProviderTwitter::ApiError do
```
hartsick, line 104: same comment here about not wanting to return an error from parse_data, though we do want to check that we send to Sentry

test/models/parser/twitter_profile_test.rb
```diff
@@ -71,62 +60,123 @@ def empty_doc
...
+        Parser::TwitterProfile.new('https://twitter.com/fake_user').parse_data(empty_doc)
+     end
       assert_equal 1, sentry_call_count
-    end
-    assert_match /Twitter::Error/, data['error']['message']
```
hartsick, line 88: I think we want to keep this (in particular the error message check), and just test whatever fields are set on error case.

test/models/parser/twitter_profile_test.rb
```diff
@@ -71,62 +60,123 @@ def empty_doc
...
+  test "should store raw data of profile returned by Twitter API" do
+    stub_profile_lookup.returns(twitter_profile_response_success)
+    
+    data = Parser::TwitterProfile.new('https://www.twitter.com/fake_user').parse_data(empty_doc)
+    
+    assert_not_nil data['raw']['api']
+    assert !data['raw']['api'].empty?
+  end  
+  
+  test "returns data even if an error is returned" do
```
hartsick, line 137: **question** does twitter return 200 and then still return an errored json response? if so, we have a few error cases (writing this out mostly because I only just realized, so that lack of previous understanding may be present in my previous comments):
1. 200 response with error in json (which you've handled)
1. non-200 response
1. exception in Net::HTTP

this seems like a good test for the first case.

test/models/parser/twitter_profile_test.rb
```diff
@@ -71,62 +60,123 @@ def empty_doc
...
+    data = Parser::TwitterProfile.new('https://twitter.com/fake_user').parse_data(empty_doc)
+    
+    assert_not_nil data['error']
+    assert_equal 'fake_user', data['external_id']
+    assert_equal 'https://twitter.com/fake_user', data['url']
   end
 
-  test "should store raw data of profile returned by Twitter API" do
-    Twitter::REST::Client.any_instance.stubs(:user).returns(fake_twitter_user)
+  test "should remove line breaks from Twitter profile description" do
```
hartsick, line 147: great test

test/models/parser/twitter_profile_test.rb
```diff
@@ -71,62 +60,123 @@ def empty_doc
...
-  test "should throw Pender::Exception::ApiLimitReached when Twitter::Error::TooManyRequests is thrown" do
-    Twitter::REST::Client.any_instance.stubs(:user).raises(Twitter::Error::TooManyRequests)
+  test "should parse tweet url with special chars, and strip them" do
+    stub_profile_lookup.returns(twitter_profile_response_success)
 
-    assert_raises Pender::Exception::ApiLimitReached do
-      Parser::TwitterProfile.new('https://twitter.com/fake-account').parse_data(empty_doc)
-    end
+    parser = Parser::TwitterProfile.new('https://0.twitter.com/fake_user')
+    data = parser.parse_data(empty_doc)
```
hartsick, line 159: small thing, but I think it helps with understandability - we don't need to assign the data variable since we don't use its result. can just do `parser.parse_data(empty_doc)` to get the side effects, but still test parser

app/models/concerns/provider_twitter.rb
hartsick, line 1: really great job!!!

hartsick (commented):
Manu, this is great so far! Left some comments, including some suggestions about how to handle and test error cases

app/models/concerns/provider_twitter.rb
vasconsaurus: honestly, I'm not sure. I tried reading on this, running some tests, and taking a look at how we do this in other parts of the code, but I'm still not sure about what is the best way to deal with this.

app/models/parser/twitter_profile.rb
vasconsaurus: we had a test that checked if any whitespaces were removed. So I assumed this was something we saw before and wanted to address.

app/models/concerns/provider_twitter.rb
hartsick: I think it's probably useful to re-raise, but definitely no straight answer. it depends on how we use this method, and where we want to raise errors (and effectively stop code execution within the closest rescue block we're in) and fail vs continue. I think in this case it would make sense to suppress the error (just return nil without re-raising) and just report the issue to Sentry since it's not a critical error - just a failure to fetch a particular attribute. I think that's why we had it swallowing the error previously, and just reported to Sentry.

If we did want to re-raise, we'd probably want to raise a different error than we're raising above it - the reason here being that if we got a non-successful error code then right now the error we would get is ApiError.new(ApiError: <message>). Since those two cases mean different things to us (non-successful response code vs any error relating to accessing the API, which could include things like timeouts or a poorly formed response body), we should have different error classes.

spec/requests/api/medias_spec.rb
vasconsaurus: I deleted because I think this is testing behavior that no longer exists. I could totally be misunderstanding, but I think this error handling came from the Twitter gem.

So I just deleted this. I'm not sure if there is a way to substitute this?

test/models/parser/page_item_test.rb
vasconsaurus: I looked through this one again and it's working, we are still testing the author_url construction. 😁

test/models/parser/page_item_test.rb
vasconsaurus: I went through this one again and I understood it better. When we get the twitter_metadata it looks  for the `twitter:creator` and tries to construct the author_url, if there isn't a twitter:creator it shouldn't error.

So I changed it to this
```
test "shouldn't error when cannot get twitter author url" do
    
  Parser::PageItem.stubs(:twitter_author_url).returns(nil)
    
  data = Parser::PageItem.new('https://random-page.com/page-item').parse_data(empty_doc)

  assert data['error'].nil?
end
```

app/models/parser/base.rb
```diff
@@ -166,13 +157,7 @@ def get_twitter_metadata
...
     def twitter_author_url(username)
       return if bad_username?(username)
-      begin
-        twitter_client.user(username)&.url&.to_s
-      rescue Twitter::Error => e
-        PenderSentry.notify(e, url: url, username: username)
-        Rails.logger.warn level: 'WARN', message: "[Parser] #{e.message}", username: username, error_class: e.class
-        nil
-      end
+        "https://twitter.com/" + username.gsub("@","")
```
caiosba, line 160: Minor: Please delete these leading spaces (we need a linter :sweat_smile:)

caiosba (approved):
Fantastic work @vasconsaurus ! Thanks for addressing the comments from the previous review. I like the fact that the unit and integration tests are separated for the Twitter parsers.

PR #374, opened 2023-08-14
3571 - Update unavailable facebook items tests

vasconsaurus:
## Description

Sometimes facebook items are unavailable. Before we didn’t get any response in those cases, now we get a page letting us know the page is not available. With this we are now able to create an embed.

We want to keep this new behavior, so we only need to update the tests.

**What the embed looks like**
Screenshot 2023-08-14 at 16 03 41

**On check**
Screenshot 2023-08-14 at 16 04 12

References: 3571

## How has this been tested?

Updated the failing tests.

### DIFF
--- a/test/models/parser/facebook_item_test.rb
+++ b/test/models/parser/facebook_item_test.rb
@@ -21,7 +21,7 @@ class FacebookItemIntegrationTest < ActiveSupport::TestCase
     assert !data['published_at'].blank?
   end
 
-  test "should set title to URL and error if crowdtangle fails" do
+  test "should set title to URL if an item is unavailable" do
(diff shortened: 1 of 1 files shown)


vasconsaurus:
@caiosba should we use this ticket as an opportunity to split facebook's integration and unit tests, or should we open a different issue to work on those?

caiosba:
> @caiosba should we use this ticket as an opportunity to split facebook's integration and unit tests, or should we open a different issue to work on those?

@vasconsaurus , since tests are currently broken, let's focus in just fixing them in the scope of this work.

caiosba (approved):
Good to go if tests pass!

PR #375, opened 2023-08-15
3227 - Remove vendorized jquery

vasconsaurus:
## Description

We got a warning on `dependabot` about a vulnerability involving our current version of `jquery`, so we needed to update it. After looking in our codebase it seems the only place we were using `jquery` was deleted ([PR 368](https://github.com/meedan/pender/pull/368)). So I removed it, instead of updating.

The file where `jquery` was used is `embed.js`, the important functionality there that should not break is the resizing of embeds. 

References: 3227, 3063

## How has this been tested?

I added different links to check and looked in check-web if they resized correctly. 

When I first tried this, there was some weirdness with the facebook embed, however, I'm not able to recreate it anymore. In my local environment all the facebook embeds have been returning the unavailable message.

I still tested with facebook (even if the embed is the unavailable message), a regular page, telegram, tiktok and twitter. They all seem to be working (Twitter seemed broken for a brief second, but a hard refresh fixed it 😅) 

In think for this one it would be important for another dev to run this locally and see if it is indeed working fine.


public/javascripts/embed.js
```diff
@@ -3,80 +3,78 @@
 // This happens on the child window
 var Pender = {};
 
-(function($) {
```
caiosba, line 6: Question: If I understand correctly, the only usage of jQuery is this `$` passed as a parameter, which is not even used, right, @vasconsaurus ? If so, I suggest we just delete the `$` and keep the rest as is (keep the `function() {`).

caiosba: The form `(function(){...})()` is a function literal that is executed immediately. What this means is that you have defined a function and you are calling it immediately.

This form is useful for information hiding and encapsulation since anything you define inside that function remains local to that function and inaccessible from the outside world. So, mostly, encapsulation.

caiosba (approved):

PR #376, opened 2023-08-17
3386 - make sure oembed_json is a Hash

vasconsaurus:
## Description

We sometimes get an error when parsing the `oembed`: `TypeError: no implicit conversion of String into Integer`

This happens with similar specific urls, all weirdly named urls with `.top` domain. Those sites return '404' on their `response.body`, when we `JSON.parse` it, it returns an integer, which breaks things. 

This was fixed by returning early if the value is not a `Hash`.

References: 3386

## How has this been tested?

Tested parsing some of the urls that were causing this error. (there are some examples on the ticket)

### DIFF
--- a/app/models/oembed_item.rb
+++ b/app/models/oembed_item.rb
@@ -42,11 +42,11 @@ class OembedItem
 
   def parse_oembed_response(response)
     return if response.nil? || response.body.blank?
-
     oembed_json = {}
     begin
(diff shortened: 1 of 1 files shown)


caiosba (approved):
Pretty straightforward :)

vasconsaurus:
Two facebook tests failed. I think there is some flakiness going on with the facebook tests. I don't think the failures are related to this change.

PR #381, opened 2023-08-24
3644 - identify facebook login pages

vasconsaurus:
## Description

We don't want the title and description to be set to 'log in to facebook'.
The title should be the url and description should be an empty string.

We were having two issues:
1. the story link was being parsed by the page parser, so we updated the regex
2. the facebook parser was being redirected to a login page, which meant that the metadata we used for the title and description was 'log in...'. Now we check one final time and update the title if needed.

References: 3644

## How has this been tested?

1. Added `assert Parser::FacebookItem.match?('https://m.facebook.com/story.php?story_fbid=pfbid0213Dz5MyduLTHpELPoRmop9E7zj3Ed163P7djxSWbkfvaMSBrjNYTY9BFx6h7i3zWl&id=100054495283578').is_a?(Parser::FacebookItem)` to  test "matches known URL patterns, and returns instance on success" (`rails test test/models/parser/facebook_item_test.rb:258`) 
2. Added a new test to check that the `title` and `description` are set correctly when the `og:title` returns 'log in or sign up to view', 'log into facebook' or 'log in to facebook' (`rails test test/models/parser/facebook_item_test.rb:475`) 

There are two failures in the tests:
1. "should not change url when redirected to login page" (`rails test test/models/parser/facebook_item_test.rb:59`) is currently failing, but it's also failing on develop. I'll deal with that issue on a separate ticket.
2. the other one is a kwai test, so that is unrelated to this PR.

### DIFF
--- a/app/models/concerns/provider_facebook.rb
+++ b/app/models/concerns/provider_facebook.rb
@@ -101,7 +101,7 @@ module ProviderFacebook
     title = get_page_title(html_page)
     return unless title
     return if NONUNIQUE_TITLES.include?(title.downcase)
-
+    
     title
   end
 end
--- a/app/models/concerns/provider_facebook.rb
+++ b/app/models/concerns/provider_facebook.rb
@@ -101,7 +101,7 @@ module ProviderFacebook
     title = get_page_title(html_page)
     return unless title
     return if NONUNIQUE_TITLES.include?(title.downcase)
-    
+
     title
   end
 end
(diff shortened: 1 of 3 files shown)


app/models/concerns/provider_facebook.rb
```diff
@@ -101,7 +101,7 @@ def get_unique_facebook_page_title(html_page)
     title = get_page_title(html_page)
     return unless title
     return if NONUNIQUE_TITLES.include?(title.downcase)
-
+
```
caiosba, line 104: Trailing space (not your fault, we need a linter :smiling_face_with_tear: )

caiosba (approved):
Looks good! I tested locally and it worked as expected for a URL that was failing before:

Captura de tela de 2023-08-30 09-56-45

PR #382, opened 2023-08-31
3644 - facebok identify login pages refactor

vasconsaurus:
## Description

I had added a final check inside `facebook_item`, but realized that it would be better to update how the `provider_facebook` checks if a page is a login page, so we check only once instead of doing it twice.
This is a refactor of the work in this PR: [3644 - identify facebook login pages #381](https://github.com/meedan/pender/pull/381). 

References: 3644

## How has this been tested?

`rails test test/models/parser/facebook_profile_test.rb:476`

I thought this was pretty straightforward, but found a small issue while working on this: 
We had a couple tests that wanted opposite things ("should add login required error and return empty html and description" / "should return html even when FB url is private"), they were both passing because we were sometimes failing to check for login pages.

The behavior we want is to return the `data[:html]` and to try to create the embed even if a login or private page:
- sometimes even if the link redirects to a login, the embed works
- even if the embed shows 'log into facebook', that might be useful as it will make it clear to the user that it's not a Check error

So I updated the tests and code accordingly.

### DIFF
--- a/app/models/parser/facebook_item.rb
+++ b/app/models/parser/facebook_item.rb
@@ -52,7 +52,7 @@ module Parser
         set_data_field('object_id', grabber.post_id)
         set_data_field('uuid', grabber.uuid)
         set_data_field('external_id', grabber.uuid)
-
+        
         @parsed_data['raw']['crowdtangle'] = get_crowdtangle_data(parsed_data['uuid']) || {}
         if has_valid_crowdtangle_data?
           crowdtangle_data = format_crowdtangle_result(parsed_data['raw']['crowdtangle'])
@@ -61,7 +61,7 @@ module Parser
           @parsed_data.merge!(crowdtangle_data)
         else
           og_metadata = get_opengraph_metadata.reject{|k,v| v.nil?}
-
+          
           if should_use_markup_title?
             set_data_field('title', get_unique_facebook_page_title(doc))
           else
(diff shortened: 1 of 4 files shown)


vasconsaurus:
**Note on the Travis failed tests**
We are only getting unrelated (expected 🥲) failures: the kwai parser test that started to fail again after working for a while, and the unrelated facebook test that will be dealt with on a different ticket.

app/models/parser/facebook_item.rb
```diff
@@ -52,7 +52,7 @@ def parse_data_for_parser(doc, original_url, _jsonld_array)
         set_data_field('object_id', grabber.post_id)
         set_data_field('uuid', grabber.uuid)
         set_data_field('external_id', grabber.uuid)
-
+
```
caiosba, line 55: Trailing space

app/models/parser/facebook_item.rb
```diff
@@ -61,7 +61,7 @@ def parse_data_for_parser(doc, original_url, _jsonld_array)
           @parsed_data.merge!(crowdtangle_data)
         else
           og_metadata = get_opengraph_metadata.reject{|k,v| v.nil?}
-
+
```
caiosba, line 64: Trailing space

caiosba (approved):

PR #385, opened 2023-09-05
3653 - extract username correctly when there is a query

vasconsaurus:
## Description
Sometimes a profile url might come with a query, e.g: `https://twitter.com/username?ref_src=blah`. 
We need to make sure we are getting only the username to make the request to the twitter api endpoint. To do that I updated the regex inside `twiter_profile.rb`.

References: 3653

## How has this been tested?

I added two new tests to the test "matches known URL patterns, and returns instance on success" (`rails test test/models/parser/twitter_profile_test.rb:41`).

I also added a new test that checks for this specific scenario "should parse tweet profile with a query on the url" (`rails test test/models/parser/twitter_profile_test.rb:203`).

### DIFF
--- a/app/models/parser/twitter_profile.rb
+++ b/app/models/parser/twitter_profile.rb
@@ -9,8 +9,10 @@ module Parser
 
       def patterns
         [
-          /^https?:\/\/(www\.)?twitter\.com\/(?<username>[\w\d]+)(\??.*)$/,
-          /^https?:\/\/(0|m|mobile)\.twitter\.com\/(?<username>[\w\d]+)(\??.*)$/
+          /^https?:\/\/(www\.)?twitter\.com\/(?<username>[\w\d]+)(\?+.*)$/,
+          /^https?:\/\/(0|m|mobile)\.twitter\.com\/(?<username>[\w\d]+)(\?+.*)$/,
+          /^https?:\/\/(www\.)?twitter\.com\/(?<username>[^\/]+)$/,
+          /^https?:\/\/(0|m|mobile)\.twitter\.com\/(?<username>[^\/]+)$/
         ]
       end
     end
--- a/app/models/parser/twitter_profile.rb
+++ b/app/models/parser/twitter_profile.rb
@@ -9,8 +9,8 @@ module Parser
 
       def patterns
         [
-          /^https?:\/\/(www\.)?twitter\.com\/(?<username>[\w\d]+)\?*[^\/]+$/,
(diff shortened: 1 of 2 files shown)


vasconsaurus:
The travis errors are the expected ones 🥲 (kwai and facebook, still need to deal with them)

test/models/parser/twitter_profile_test.rb
```diff
@@ -193,4 +199,13 @@ def stub_profile_lookup
...
   end
+
+  test "should parse tweet profile with a query on the url" do
+    stub_profile_lookup.returns(twitter_profile_response_success)
+
+    data = Parser::TwitterProfile.new('https://www.twitter.com/fake_user?ref_src=twsrc%5Etfw').parse_data(empty_doc)
+
+    assert_equal 'fake_user', data['external_id']
+    assert_equal '@fake_user', data['username']
+  end
```
caiosba, line 210: Nice test!

app/models/parser/twitter_profile.rb
```diff
@@ -9,6 +9,8 @@ def type
 
       def patterns
         [
+          /^https?:\/\/(www\.)?twitter\.com\/(?<username>[\w\d]+)(\?+.*)$/,
+          /^https?:\/\/(0|m|mobile)\.twitter\.com\/(?<username>[\w\d]+)(\?+.*)$/,
           /^https?:\/\/(www\.)?twitter\.com\/(?<username>[^\/]+)$/,
           /^https?:\/\/(0|m|mobile)\.twitter\.com\/(?<username>[^\/]+)$/
```
caiosba, line 15: Question: These two regular expressions are still too broad and will keep matching things we don't want, right? But it can be a separate ticket, where we follow this: https://help.twitter.com/en/managing-your-account/twitter-username-rules#error
```
Your username cannot be longer than 15 characters. Your name can be longer (50 characters) or shorter than 4 characters, but usernames are kept shorter for the sake of ease.

A username can only contain alphanumeric characters (letters A-Z, numbers 0-9) with the exception of underscores, as noted above. Check to make sure your desired username doesn't contain any symbols, dashes, or spaces.
```

caiosba (approved):

app/models/parser/twitter_profile.rb
caiosba: Let's leave it as is for now, and refine as we notice those cases (like we're handling in this ticket).

PR #387, opened 2023-09-07
3723 – Fix Kwai parser

vasconsaurus:
## Description

Our Kwai test has been failing for a while, since it had been flaky before we assumed it was just flaky again. However, it seems that the parsing is indeed broken: we search the doc for '.info .title' and for '.name', and both of those are returning nil.

We can get data from the jsonld script. Besides the `name`(we can get that from `creator name`), it doesn't return the same exact data as before. We can use `description` or `transcription` for `data['description']`. For `title` we don't have a good substitute, so we are just using the default fallback from `Media` (`url`).

We are leaving the old behavior and adding the data from the jsonld script as a fallback. Kwai has changed their HTML and changed it back again before. So we think it's worth keeping the old behavior, at least for now.

References: 3723, 3716

## How has this been tested?

I updated the integration test to use the new values we are fetching (`rails test test/models/parser/kwai_test.rb:4`), and added two unit tests:
- make sure we are getting author_name and description from json+ld successfully (`rails test test/models/parser/kwai_test.rb:50`)
- make sure the title uses the url as a fallback (`rails test test/models/parser/kwai_test.rb:61`)

### DIFF
--- a/app/models/parser/kwai_item.rb
+++ b/app/models/parser/kwai_item.rb
@@ -19,7 +19,7 @@ module Parser
       handle_exceptions(StandardError) do
         jsonld = (jsonld_array.find{|item| item.dig('@type') == 'VideoObject'} || {})
 
-        title = get_kwai_text_from_tag(doc, '.info .title') || jsonld.dig('name')
+        title = get_kwai_text_from_tag(doc, '.info .title') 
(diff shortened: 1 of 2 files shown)


caiosba (approved):
Thanks! It aligns with what we discussed in the ticket.

PR #388, opened 2023-09-11
3631: Refactor Facebook testing

vasconsaurus:
## Description

We are updating the Facebook profile and item parsers' test.

**Context**
We have two issues:
1. We want to be able to choose when to run integration tests.
2. We have too many integration tests.

To fix this we need to go through all the parsers and:
- Split the integration from the unit tests, we now have a folder for them (test > integration > parsers).
- Go through the unit tests and make sure they are not making live http requests.
- Go through the integration tests, see which one of those should be integration tests (keep those at 2, 3 tests at the most), and refactor the rest to be unit tests. (ideally we want one to test a successful request and one to test a failure)

**Files**
- `facebook_item_ids_grabber_test.rb`
- `facebook_item_test.rb`
- `facebook_profile_test.rb`

References: 3631, 3629

`facebook_item_test` was trickier, because we use crowdtangle, so there was more mocking, stubbing and test cases in that file. `facebook_item_ids_grabber_test.rb` didn't need any updates.


caiosba (approved):
Amazing work, @vasconsaurus !

PR #392, opened 2023-09-18
Make regex on twitter_profile less broad

vasconsaurus:
## Description

We were looking for anything that wasn’t ‘/', we want to make that less broad by only looking for [what actually can be in a username](https://help.twitter.com/en/managing-your-account/twitter-username-rules#error).

This way something like a search won’t be wrongly parsed as an user. It’s better to have the page parser try to parse that than the twitter profile parser. We know the twitter parser will fail, while the page parser might be able to return something.

References: 3717, 3740

## How has this been tested?

Added a new test "does not match patterns with usernames that are not permitted by twitter"
`rails test test/models/parser/twitter_profile_test.rb:71`

Also ran "matches known URL patterns, and returns instance on success" and "does not match pages that should be parsed by pages" to make sure nothing broke.
`rails test test/models/parser/twitter_profile_test.rb:41`
`rails test test/models/parser/twitter_profile_test.rb:61`

### DIFF
--- a/app/models/parser/twitter_profile.rb
+++ b/app/models/parser/twitter_profile.rb
@@ -9,7 +9,7 @@ module Parser
 
       def patterns
         [
-          /^https?:\/\/((0|m|mobile|www)\.)?twitter\.com\/(?<username>[\w]{4,15})(\?.*)?[\/]*$/
+          /^https?:\/\/((0|m|mobile|www)\.)?twitter\.com\/(?<username>[\w]{4,15})[\/]*(\?.*)?[\/]*$/
         ]
       end
     end
--- a/test/models/parser/twitter_profile_test.rb
+++ b/test/models/parser/twitter_profile_test.rb
@@ -50,16 +50,18 @@ class TwitterProfileUnitTest < ActiveSupport::TestCase
     # Profile with query
     match_three = Parser::TwitterProfile.match?('https://twitter.com/username?ref_src=twsrc%5Etfw')
     assert_equal true, match_three.is_a?(Parser::TwitterProfile)
+    match_four = Parser::TwitterProfile.match?('https://twitter.com/username/?t=1')
+    assert_equal true, match_four.is_a?(Parser::TwitterProfile)
 
     # Mobile patterns
-    match_four = Parser::TwitterProfile.match?('https://0.twitter.com/username')
-    assert_equal true, match_four.is_a?(Parser::TwitterProfile)
-    match_five = Parser::TwitterProfile.match?('https://m.twitter.com/username')
+    match_five = Parser::TwitterProfile.match?('https://0.twitter.com/username')
     assert_equal true, match_five.is_a?(Parser::TwitterProfile)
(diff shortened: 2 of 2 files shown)


vasconsaurus:
the test failing is a facebook one, that is working in the `3631-refactor-facebook-testing` branch (where it has already been updated)

app/models/parser/twitter_profile.rb
```diff
@@ -11,8 +11,8 @@ def patterns
         [
           /^https?:\/\/(www\.)?twitter\.com\/(?<username>[\w\d]+)(\?+.*)$/,
           /^https?:\/\/(0|m|mobile)\.twitter\.com\/(?<username>[\w\d]+)(\?+.*)$/,
-          /^https?:\/\/(www\.)?twitter\.com\/(?<username>[^\/]+)$/,
-          /^https?:\/\/(0|m|mobile)\.twitter\.com\/(?<username>[^\/]+)$/
+          /^https?:\/\/(www\.)?twitter\.com\/(?<username>[\w\d]{4,15})$/,
+          /^https?:\/\/(0|m|mobile)\.twitter\.com\/(?<username>[\w\d]{4,15})$/
```
caiosba, line 15: Request: The two first regular expressions can still match usernames with less than 4 characters or more than 15 characters, so we need to fix them.

Tip: `\w` already includes digits, so `\w\d` can be replaced by just `\w`.

Suggestion: Combine all four regular expressions into a single one... something like this (untested):
```
/^https?:\/\/((0|m|mobile|www)\.)?twitter\.com\/(?<username>[\w]{4,15})(\?.*)?$/
```

Question: What about trailing slashes? Are they stripped by normalization or some previous step?

caiosba (changes requested):

app/models/parser/twitter_profile.rb
vasconsaurus: > Question: What about trailing slashes? Are they stripped by normalization or some previous step?

They were not, I updated the regex to also take trailing slashes into account

app/models/parser/twitter_profile.rb
```diff
@@ -9,10 +9,7 @@ def type
 
       def patterns
         [
-          /^https?:\/\/(www\.)?twitter\.com\/(?<username>[\w\d]+)(\?+.*)$/,
-          /^https?:\/\/(0|m|mobile)\.twitter\.com\/(?<username>[\w\d]+)(\?+.*)$/,
-          /^https?:\/\/(www\.)?twitter\.com\/(?<username>[^\/]+)$/,
-          /^https?:\/\/(0|m|mobile)\.twitter\.com\/(?<username>[^\/]+)$/
+          /^https?:\/\/((0|m|mobile|www)\.)?twitter\.com\/(?<username>[\w]{4,15})(\?.*)?[\/]*$/
```
caiosba, line 12: Good job! Minor change, but needed: `/` should come before the query parameters, so it can match something like this: `https://twitter.com/foo_bar/?t=1`.

caiosba (changes requested):

caiosba (approved):

PR #393, opened 2023-09-21
notify Sentry about RetryLimitHit only from sidekiq

vasconsaurus:
## Description

We were notifying Sentry both through `Media.give_up` and through
configuration inside `03_sidekiq`. We are keeping only the one in `03_sidekiq.rb`.
It’s the most specific error and the right place.

In order to try to get more information, we are replacing
`Sentry.capture_exception(ex) by PenderSentry.notify(ex, { job: job })`

References: 3706, 3757, 3707

## How has this been tested?

I tried writing a test to make sure teh Archiver isn't notifying Sentry anymore:
test "MediaArchiver should not notify Sentry when the worker hits the maximum number of retries"
`rails test test/models/archiver_test.rb:746`

I ran a bunch of tests to make sure the test is working, but I'm not sure about it: if it makes sense the way it's written, if it's even needed.

### DIFF
--- a/config/initializers/03_sidekiq.rb
+++ b/config/initializers/03_sidekiq.rb
@@ -12,9 +12,9 @@ if File.exist?(file)
 
     config.death_handlers << ->(job, ex) do
       if ex.is_a?(Pender::Exception::RetryLater)
-        ex = Pender::Exception::RetryLimitHit.new(ex, {job: job})
+        ex = Pender::Exception::RetryLimitHit.new(ex)
       end
-      Sentry.capture_exception(ex)
+      PenderSentry.notify(ex, {job: job})
     end
   end
 
--- a/config/initializers/03_sidekiq.rb
+++ b/config/initializers/03_sidekiq.rb
@@ -12,7 +12,7 @@ if File.exist?(file)
 
     config.death_handlers << ->(job, ex) do
       if ex.is_a?(Pender::Exception::RetryLater)
-        ex = Pender::Exception::RetryLimitHit.new(ex)
+        ex = Pender::Exception::RetryLimitHit.new(ex, {job: job})
       end
       Sentry.capture_exception(ex)
     end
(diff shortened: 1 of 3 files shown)


config/initializers/03_sidekiq.rb
```diff
@@ -12,7 +12,7 @@
 
     config.death_handlers << ->(job, ex) do
       if ex.is_a?(Pender::Exception::RetryLater)
-        ex = Pender::Exception::RetryLimitHit.new(ex)
+        ex = Pender::Exception::RetryLimitHit.new(ex, {job: job})
```
caiosba, line 15: The exception doesn't take a second argument, it should remain like it was:
```
ex = Pender::Exception::RetryLimitHit.new(ex)
```

config/initializers/03_sidekiq.rb
```diff
@@ -12,7 +12,7 @@
...
       end
       Sentry.capture_exception(ex)
```
caiosba, line 17: What changes is here. We need to be able to send additional information along with the Sentry notification. I don't know if `Sentry.capture_exception` does that. If not, I suggest replacing this line by:
```
PenderSentry.notify(ex, { job: job })
```

caiosba (changes requested):

caiosba (approved):
Looks good! Good to go once tests pass. The new test looks good too, just one question: is that `ensure WebMock.disable!` needed? Isn't that handled by the `teardown` method?

vasconsaurus:
> Looks good! Good to go once tests pass. The new test looks good too, just one question: is that `ensure WebMock.disable!` needed? Isn't that handled by the `teardown` method?

Not sure. Since the other tests inside this file had it (the `ensure WebMock.disable!`) I thought it was best to keep it.

In our tests where we don't have the `ensure WebMock.disable!` we are using the `isolated_teardown` from `test_helper`. Which I don't think is the case here.

What do you think?

caiosba:
> > Looks good! Good to go once tests pass. The new test looks good too, just one question: is that `ensure WebMock.disable!` needed? Isn't that handled by the `teardown` method?
> 
> Not sure. Since the other tests inside this file had it (the `ensure WebMock.disable!`) I thought it was best to keep it.
> 
> In our tests where we don't have the `ensure WebMock.disable!` we are using the `isolated_teardown` from `test_helper`. Which I don't think is the case here.
> 
> What do you think?

OK, sounds good to keep it as is for now.

PR #397, opened 2023-09-27
3718 – Update how we retry tests

vasconsaurus:
## Description

When running locally we might not want the tests to retry, to do this
we would need to comment out the retry lines. We want to make this easier
on the developer

Converting the env variable to an integer because ENV[] on ruby returns a string
https://ruby-doc.org/core-3.0.0/ENV.html

References: 3718

## How has this been tested?

Changed a test to make it fail. 
Ran tests locally without setting `TEST_RETRY_COUNT`, test failed and didn't retry.
Set the env var inside the container (`export TEST_RETRY_COUNT=5`) and ran the tests again, it retried 5 times.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,8 +10,8 @@ before_install:
 before_script:
 - docker-compose exec pender test/setup-parallel
 script:
-- docker-compose exec pender bundle exec rake "parallel:test[3]"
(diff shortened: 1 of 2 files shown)


caiosba (approved):
Nice, Manu!

PR #399, opened 2023-10-03
3634 – Refactor Kwai testing

vasconsaurus:
## Description

This is part of a larger project to improve our testing.

**Context**
We have two issues:

- We want to be able to choose when to run integration tests.
- We have too many integration tests.

To fix this we need to go through all the parsers and:

- Split the integration from the unit tests, we now have a folder for them (test > integration > parsers).
- Go through the unit tests and make sure they are not making live http requests.
- Go through the integration tests, see which one of those should be integration tests (keep those at 2, 3 tests at the most), and refactor the rest to be unit tests. (ideally we want one to test a successful request and one to test a failure)

**Notes on the Kwai tests**

This one was pretty straight forward:
- I moved integration test to it's own file
- Created a new integration test that makes sure it returns data even if it fails
- No clean up was needed

References: 3634, 3629

### DIFF
--- a/test/integration/parsers/kwai_test.rb
+++ b/test/integration/parsers/kwai_test.rb
@@ -1,0 +1,27 @@
+require 'test_helper'
+
+class KwaiIntegrationTest < ActiveSupport::TestCase
+  test "should parse Kwai URL" do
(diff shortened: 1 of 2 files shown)


caiosba (approved):
Nice and clean!

PR #403, opened 2023-10-05
Fixing cache when ID changes

caiosba:
## Description

The URL ID, used as a cache key, is calculated in runtime, at different times, when writing and when reading. This means that the cache is written to a key but read from another one. This is a pretty edge case, but this PR turns it safer and more consistent.

Reference: CV2-3749.

## How has this been tested?

TDD. I added a test that first reproduced the issue. The test passed after the fix was implemented.

## Checklist

- [x] I have performed a self-review of my own code
- [x] I have added unit and feature tests, if the PR implements a new feature or otherwise would benefit from additional testing
- [x] I considered secure coding practices when writing this code. Any security concerns are noted above.
- [x] My changes generate no new warnings

### DIFF
--- a/test/controllers/medias_controller_test.rb
+++ b/test/controllers/medias_controller_test.rb
@@ -690,7 +690,7 @@ class MediasControllerTest < ActionController::TestCase
   test "should refresh cache even if ID changes" do
     WebMock.enable!
     WebMock.disable_net_connect!(allow: [/minio/])
-    WebMock.stub_request(:get, /.*/).to_return(status: 200, body: '<html><title>Test</title></html>')
+    WebMock.stub_request(:post, /safebrowsing\.googleapis\.com/).to_return(status: 200, body: '{}')
     Media.stubs(:get_id).returns('foo', 'bar', 'foo', 'bar')
     Pender::Store.current.write('foo', :json, { title: 'Meedan 1' })
     Pender::Store.current.write('bar', :json, { title: 'Meedan 2' })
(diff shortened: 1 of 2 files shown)


test/controllers/medias_controller_test.rb
```diff
@@ -686,6 +686,22 @@ def setup
     assert_equal url, JSON.parse(@response.body)['data']['title']
     assert_equal Lapis::ErrorCodes::const_get('DUPLICATED'), JSON.parse(@response.body)['data']['error']['code']
   end
+
+  test "should refresh cache even if ID changes" do
+    WebMock.enable!
+    WebMock.disable_net_connect!(allow: [/minio/])
+    Media.stubs(:get_id).returns('foo', 'bar', 'foo', 'bar')
```
vasconsaurus, line 694: @caiosba if/when you have the time, would you mind explaining how line 693 works again?

caiosba: Sure! It basically means that each time the stubbed method (in this case, `get_id`) is called during that test, the returned value will be different. So, first time it will be `foo`, second time it will be `bar`.

vasconsaurus: why `.returns('foo', 'bar', 'foo', 'bar')` and not `.returns('foo', 'bar')` ?

caiosba: Because it runs four times during that test, 2 in the controller and 2 in the model. I can show you when we pair and explain.

PR #405, opened 2023-10-11
3843 – Unrelated instagram links from the same page being matched

vasconsaurus:
## Description

One of our users saw a bug where some unrelated items were being matched. 

This happened because even though they linked to different urls their data was the same (the title for all of them were the instagram home url). They were also being parsed by page and not instagram item parser.

References: 3843

## How has this been tested?

Added a live test (`test "should parse Instagram item when the final url is instagram.com"`) and a stubbed test (`test "should return url as title when redirected to instagram main page"`).

### DIFF
--- a/app/models/parser/instagram_item.rb
+++ b/app/models/parser/instagram_item.rb
@@ -3,6 +3,7 @@ module Parser
     include ProviderInstagram
 
     INSTAGRAM_ITEM_URL = /^https?:\/\/(www\.)?instagram\.com\/(p|tv|reel)\/([^\/]+)/
+    INSTAGRAM_HOME_URL = /^https?:\/\/(www\.)?instagram\.com\/?$/
     
     class << self
       def type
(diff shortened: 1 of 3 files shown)


app/models/parser/instagram_item.rb
```diff
@@ -3,21 +3,22 @@ class InstagramItem < Base
...
-        [INSTAGRAM_ITEM_URL]
+        [INSTAGRAM_ITEM_URL, INSTAGRAM_HOME_URL]
       end
     end
     
     private
 
     # Main function for class
-    def parse_data_for_parser(doc, _original_url, _jsonld_array)
+    def parse_data_for_parser(doc, original_url, _jsonld_array)
```
caiosba, line 21: @vasconsaurus , please delete this trialing space before merging :smile:

caiosba (approved):

app/models/parser/instagram_item.rb
vasconsaurus: done! 😁

PR #406, opened 2023-10-16
Update when to archive to archive.org and perma.cc

vasconsaurus:
## Description

We realized we needed to update how we deal with when to archive, and that we should check for those conditions only in Media, `media_perma_cc_archiver` and `media_archive_org_archiver` should only worry about archiving to their respective site.

References: 3707

## How has this been tested?

The tests are currently passing. For perma_cc: I updated one(`"should not try to archive on Perma.cc if already present in cache and no refresh is requested"`) and added another (`"should try to archive on Perma.cc if already present in cache and but refresh is requested"`).

Please describe parts of the change that require extra attention during code review, for example:

- are the conditionals making sense? is there anything I missed?
- is there any logic left inside `media_perma_cc_archiver` or `media_archive_org_archiver` related to this that should be moved?

### DIFF
--- a/test/models/archiver_test.rb
+++ b/test/models/archiver_test.rb
@@ -167,19 +167,19 @@ class ArchiverTest < ActiveSupport::TestCase
     WebMock.disable!
   end
 
-  test "should not archive in any archiver if don't send or it's none" do
+  test "should not archive in any archiver if none is requested" do
     Media.any_instance.unstub(:archive_to_archive_org)
     a = create_api_key application_settings: { 'webhook_url': 'https://example.com/webhook.php', 'webhook_token': 'test' }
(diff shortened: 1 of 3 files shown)


test/models/archiver_test.rb
```diff
@@ -284,15 +283,15 @@ def quietly_redefine_constant(klass, constant, new_value)
 
   test "should archive to perma.cc and store the URL on archives if perma_cc_key is present" do
     Media.any_instance.unstub(:archive_to_perma_cc)
-
+
```
caiosba, line 286: Trailing space :P

caiosba (approved):
Great testing coverage here!

PR #410, opened 2023-10-23
Retry ArchiveOrgErrors

vasconsaurus:
## Description

We split archive org errors into two: Pender::Exception::TooManyCaptures and Pender::Exception::ArchiveOrgError. It makes sense for Pender::Exception::TooManyCaptures to not be retried, but Pender::Exception::ArchiveOrgError should. So we updated `send_to_archive_org` in `media_archive_org_archiver`.

References: 3707

## How has this been tested?

Test "should update media with error when Archive.org can't archive the url" makes sure we are getting the error data when we hit ArchiveOrgError. 
Test "when archive.org fails to archive, it should add to data the available archive.org snapshot (if available) and the error" makes sure we are getting the data (and snapshot) when we hit TooManyCaptures.
The tests here maybe should be renamed, but I'll look into that while going through the tests' refactoring.

### DIFF
--- a/app/models/concerns/media_archive_org_archiver.rb
+++ b/app/models/concerns/media_archive_org_archiver.rb
@@ -36,7 +36,7 @@ module MediaArchiveOrgArchiver
               response_body: body
             )
           else
-            raise Pender::Exception::ArchiveOrgError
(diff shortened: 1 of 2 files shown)


caiosba (approved):
Thanks Manu, this is exactly what we discussed.

PR #412, opened 2023-10-26
3887 – Better error handling in PermaCC

vasconsaurus:
## Description

We want PermaCC to have specific errors, like ArchiveOrg, so we can better track and handle errors. So I created PermaCcError, and it's also using TooManyCaptures once we hit it's limit.

Importat: While working on tests I noticed that we were raising RetryLater when we errored in PermaCC, and that was causing it to behave in a different way then expected (it wasn't updating data and notifying the webhook, that would only happen when we hit the retry limit). So we now raise PermaCcError. It's retried, updates data and notifies the webhook.

Besides that, we don't want to retry when the archiving limit is hit. So when we hit TooManyCaptures it updates data and the webhook, but does not retry.

CV2-3886

## How has this been tested?

`test "when Perma.cc fails with Pender::Exception::PermaCcError it should update media with error and retry"`
`test "when Perma.cc fails with Pender::Exception::TooManyCaptures it should update media with error and not retry"`

Not sure if it's something to pay attention, but it's something to note: I skipped the video archiving tests (since we are not using that feature anymore), hoping it would make the tests a bit faster but it didn't make much of a difference 😅 (the tests on the refactoring branch are much faster ❤️ )

### DIFF
--- a/app/models/concerns/media_perma_cc_archiver.rb
+++ b/app/models/concerns/media_perma_cc_archiver.rb
@@ -30,7 +30,15 @@ module MediaPermaCcArchiver
           data = { location: 'http://perma.cc/' + body['guid'] }
           Media.notify_webhook_and_update_cache('perma_cc', url, data, key_id)
         else
-          raise Pender::Exception::PermaCcError, "(#{response.code}) #{response.message}"
+          if response&.body.include?("You've reached your usage limit")
+            PenderSentry.notify(
+              Pender::Exception::TooManyCaptures.new(response.message),
+              url: url,
+              response_body: response.body
+            )
+          else  
+            raise Pender::Exception::PermaCcError, "(#{response.code}) #{response.message}"
+          end
         end
       end
     end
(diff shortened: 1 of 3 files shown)


app/models/concerns/media_perma_cc_archiver.rb
```diff
@@ -30,7 +30,17 @@ def send_to_perma_cc(url, key_id, _supported = nil)
           data = { location: 'http://perma.cc/' + body['guid'] }
           Media.notify_webhook_and_update_cache('perma_cc', url, data, key_id)
         else
-          raise Pender::Exception::RetryLater, "(#{response.code}) #{response.message}"
+          data = { error: { message: response.message, code: Lapis::ErrorCodes::const_get('ARCHIVER_ERROR') }}
+          Media.notify_webhook_and_update_cache('perma_cc', url, data, key_id)
+          if response&.body.include?("You've reached your usage limit")
```
caiosba, line 35: It would be good to have an integration test for this case, just for us to capture if they change their message. On the other hand, this is hard because we would need to have a dedicated account to use only for this test. Hopefully we can track that in the generic `Pender::Exception::PermaCcError` exception report. So let's not change anything for now, this is good to go.

caiosba (approved):

PR #416, opened 2023-11-27
3480 – Upgrade to Sidekiq 7

vasconsaurus:
## Description

We updated to Sidekiq 7, but there were some big changes between 6 and 7. So I'm linking the guides and explaining each step that we took.

**Guides**
https://github.com/sidekiq/sidekiq/blob/main/docs/7.0-API-Migration.md
https://github.com/sidekiq/sidekiq/blob/main/docs/7.0-Upgrade.md

**Steps**

1. We had to update Redis (Redis 6.2+ is now required)
   a. We already use Redis 6.2.2 on Live and QA
   b. We need to update in Pender, but also in Check and Check-Web
   c. We need to run integration tests to make sure this doesn't break anything.
2. strict arguments
  a. Add strict argument checking [#5071] Sidekiq will now log a warning if JSON-unsafe arguments are passed to perform_async. Add Sidekiq.strict_args!(false) to your initializer to disable this warning. This warning will switch to an exception in Sidekiq 7.0. [[link]](https://github.com/sidekiq/sidekiq/blob/main/Changes.md#640)
   b. We could silence strict args, but I was able to fix the instances we had 
3. delayed extensions
   a. They were disabled in Sidekiq 5 and removed in Sidekiq 7. [[link](https://github.com/sidekiq/sidekiq/wiki/Delayed-extensions)]
   b. I replaced where we were using delayed extensions to use a Job (I used job instead of worker because of [this](https://github.com/sidekiq/sidekiq/wiki/Best-Practices#4-use-precise-terminology) & [this](https://stackoverflow.com/questions/76095172/sidekiq-worker-vs-job))
4. logs
   a. Sidekiq info logs were quite noisy when running the tests, so I configured log level to WARN [[link](https://github.com/sidekiq/sidekiq/issues/5907#issuecomment-1536457365)]
5. Redis/Travis
   a. I ran into some Travis/Redis issues, because of that some tests that depended on Redis were failing in Travis. The solution to that was specifying the distro to `jammy` inside `travis.yml`

References: 3480

## How has this been tested?

By running the usual tests and integration tests on check-web.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: minimal
+dist: jammy
 before_install:
 - openssl aes-256-cbc -K $encrypted_3491736328a2_key -iv $encrypted_3491736328a2_iv -in config/config.yml.enc -out config/config.yml -d
 - cp config/database.yml.example config/database.yml
--- a/Gemfile
+++ b/Gemfile
@@ -45,7 +45,7 @@ gem 'retryable'
 gem 'puma', '5.6.7'
 gem 'rack-cors', :require => 'rack/cors'
 gem 'rails-perftest'
-gem 'sidekiq', '< 7'
+gem 'sidekiq', '< 8'
 gem 'redis', '4.3.1'
 gem 'nokogiri', '1.14.3', require: false
 gem 'htmlentities', require: false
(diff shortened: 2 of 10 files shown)


caiosba (approved):
Nice and clean!

PR #430, opened 2024-01-12
3908 – Improve Metrics Update

vasconsaurus:
## Description

Context:
* Today, we schedule metrics to be updated for a single URL everyday for a few days. When it fails, this means that the error propagation gets exponential
* We don’t have “metrics on metrics”, so we can’t really compare the number of errors with the number of requests

Scope:
* Log every time metrics are requested, differentiating between first request and updates:
    * **In our logs we are already differentiating between first attempt and update numbers by logging the `update_number`. The first attempt is logged as `update_number: 0`. So there was no need to add extra logging.** 
* Reduce the number of times we update metrics
    * **Reduced to one update in two days.**

References: 3908

## How has this been tested?

Ran the metrics_test file and updated a test (`rails test test/models/metrics_test.rb:81`).
Also just ran the code and took a look at the logs and sidekiq.

While testing this I wasn't able to make one successful metrics request, they all failed:
```
"error_message":"(425) CUSTOM","error_class":"Pender::Exception::RetryLater"
```

### DIFF
--- a/app/models/metrics.rb
+++ b/app/models/metrics.rb
@@ -13,6 +13,7 @@ module Metrics
       613, # Calls to graph_url_engagement_count have exceeded the rate of 10 calls per 3600 seconds
     ]
 
+    # we are going to update once in 2 days
     NUMBER_OF_DAYS_TO_UPDATE = 1
 
     def schedule_fetching_metrics_from_facebook(data, url, key_id)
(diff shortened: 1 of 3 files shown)


app/models/metrics.rb
```diff
@@ -13,15 +13,16 @@ class << self
       613, # Calls to graph_url_engagement_count have exceeded the rate of 10 calls per 3600 seconds
     ]
 
-    NUMBER_OF_DAYS_TO_UPDATE = 9
+    # we are going to update once in 2 days
+    NUMBER_OF_DAYS_TO_UPDATE = 1
```
caiosba, line 17: Noting that since it's only 1, we don't need a constant and the `times` loop below, but on the other hand, let's keep some flexibility, so it's fine to keep the block.

caiosba (approved):

app/models/metrics.rb
vasconsaurus: Yup. Thought the same thing, since it's not clear what the issue with metrics is, we might want to increase it again in a not so distant future.

caiosba: Yes exactly!

PR #455, opened 2024-06-13
4518 – Create metrics class abstraction using Prometheus ruby client

vasconsaurus:
## Description

- Added the gem `prometheus-client`.
- Created a class abstraction `MetricsService`, so it's easy for us to change between prometheus, otel, or whatever else we might think is best.
- Created some very basic methods to create and increment a custom counter.
- Created one metric `media_request_total` so we can start testing things.
- We create and register custom metrics once on initialization, and then call increment wherever it's needed
- The gem adds `http_server_request_duration_seconds histogram`, `http_server_requests_total counter` and `http_server_exceptions_total counter` by default.

References: 4396, 4518

## How has this been tested?

I added some very basic tests, just to make sure I was implementing the methods correctly. Please suggests any tests you think are needed.

And tested by making requests and checking the logs on the Prometheus endpoint:

Screenshot 2024-06-13 at 16 43 51

### DIFF
--- a/lib/metrics_service.rb
+++ b/lib/metrics_service.rb
@@ -2,7 +2,7 @@ require 'prometheus/client'
 
 class MetricsService
   class << self
-    def custom_counter(name, description, labels: {})
+    def custom_counter(name, description, labels: [])
       counter = Prometheus::Client::Counter.new(
         name, 
         docstring: description,
@@ -11,14 +11,14 @@ class MetricsService
       prometheus_registry.register(counter)
     end
 
-    def increment_counter(name, labels: {})
+    def increment_counter(name, labels: [])
       counter = prometheus_registry.get(name)
       return if counter.nil?
 
       counter.increment(labels: labels)
     end
 
-    def get_counter(counter, labels: {})
+    def get_counter(counter, labels: [])
       counter.get(labels: labels)
     end
 
--- a/lib/metrics_service.rb
+++ b/lib/metrics_service.rb
@@ -1,0 +1,32 @@
+require 'prometheus/client'
+
+class MetricsService
+  class << self
+    def custom_counter(name, description, labels: {})
+      counter = Prometheus::Client::Counter.new(
+        name, 
+        docstring: description,
+        labels: labels
+        )
+      prometheus_registry.register(counter)
+    end
+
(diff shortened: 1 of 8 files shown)


vasconsaurus:
There are some weird errors in there, that I'm also getting locally. Will look into those.

vasconsaurus:
@caiosba there were some errors due to the order I added the `MetricsService.increment_counter` function. And a unrelated dropbox parser error. Those were fixed.

Now we are seeing mostly `auth failed ip_blacklisted` from facebook, instagram and tik tok, and a few other errors that are caused by the same thing. Regarding those errors:
- Should we go back to cleaning up some of the live requests we are making?
- Should I re-run the tests (maybe after lunch to give it some time), or do you think we are ok to go?

caiosba:
> @caiosba there were some errors due to the order I added the `MetricsService.increment_counter` function. And a unrelated dropbox parser error. Those were fixed.
> 
> Now we are seeing mostly `auth failed ip_blacklisted` from facebook, instagram and tik tok, and a few other errors that are caused by the same thing. Regarding those errors:
> 
> * Should we go back to cleaning up some of the live requests we are making?
> * Should I re-run the tests (maybe after lunch to give it some time), or do you think we are ok to go?

@vasconsaurus I think we can retry at least once later, but fixing these are out of scope of this ticket. Let's capture it in a new ticket of type "automated test".

caiosba (approved):

lib/metrics_service.rb
```diff
@@ -0,0 +1,32 @@
+require 'prometheus/client'
+
+class MetricsService
+  class << self
+    def custom_counter(name, description, labels: {})
+      counter = Prometheus::Client::Counter.new(
+        name,
```
skyemeedan, line 7: I think followed the pattern ( recommended on one of the sites) to include the name of the service (i.e. pender) as a label at this label so that it is easy to separate out from other similar metrics

vasconsaurus: That is a good idea! Will update it. Thanks Skye!

vasconsaurus: for now I added the label `service` to the custom metric we are creating, but we can revisit it and add to the wrapper class itself if we think it makes more sense.

PR #458, opened 2024-07-16
4520 – Add metrics to code and create Honeycomb status dashboard: parser requests [WIP]

vasconsaurus:
## Description

We are now sending metrics related to parsing requests. We are keeping track of: service_name, parser_name, parsed_host and parsing_status.

What we want to answer:
1. total of parsing requests per x amount of time
2. total of parsing requests per x amount of time grouped by partner [TBD: not possible to do right now]
3. total of parsing requests per x amount of time grouped by parser
4. total of parsing requests of an unique host, which does not have a dedicated parser, per x amount of time
5. total of errors and total of successes of our parsing requests per x amount of time

References: 4396, 4520

## How has this been tested?

By building the boards, and making sure the data is getting to Honeycomb.

### On why we are sending the data we are sending

I ended up splitting the metrics so we can get more granular data and more generic data. I did this because it was proving difficult to get the information we wanted on the Honeycomb side.

For some cases we wanted a SUM of the value send by the metric, but I think since the `otel-collector` sends the scraped data every 15 seconds we end up with a bunch of events. The SUM doesn't really show what we need. We can `zoom in` into specific points in time, and I have set the `default granularity` to match the `otel-collector`. All that helps, but doesn't get 100% what we would like.

I ended up using MAX instead of SUM. But that was an issue when we had more granular metrics and wanted to know, for example, the amount of parsing requests to the `page` parser, without breaking it up by `host`.

If we made 2 parsing requests to `page` but to different `hosts`, those would be split. And we would get the MAX 1 instead of 2.

So instead of this:
```
parser_requests{service_name="pender",parser_name="page",parsed_host="a.host",parsing_status="success"} 1.0
parser_requests{service_name="pender",parser_name="page",parsed_host="another.host",parsing_status="success"} 1.0
```

We also want this
```
parser_requests{service_name="pender",parser_name="page",parsing_status="success"} 2.0
```

### What the metrics endpoint is looking like
note: This are the metrics that are relevant to this PR.

```
# TYPE parser_requests_total counter
# HELP parser_requests_total Count parsing requests, gets the full total - does not break them by labels
parser_requests_total{service_name="pender"} 4.0

# TYPE parser_requests_success counter
# HELP parser_requests_success Count successful parsing requests
parser_requests_success{service_name="pender"} 3.0

# TYPE parser_requests_error counter
# HELP parser_requests_error Count errored parsing requests
parser_requests_error{service_name="pender"} 1.0

# TYPE parser_requests_per_parser counter
# HELP parser_requests_per_parser Count parsing requests per parser
parser_requests_per_parser{service_name="pender",parser_name="youtube",parsing_status="success"} 1.0
parser_requests_per_parser{service_name="pender",parser_name="page",parsing_status="success"} 2.0
parser_requests_per_parser{service_name="pender",parser_name="instagram",parsing_status="error"} 1.0

# TYPE parser_requests_success_per_parser counter
# HELP parser_requests_success_per_parser Count successful parsing requests per parser
parser_requests_success_per_parser{service_name="pender",parser_name="youtube"} 1.0
parser_requests_success_per_parser{service_name="pender",parser_name="page"} 2.0

# TYPE parser_requests_error_per_parser counter
# HELP parser_requests_error_per_parser Count errored parsing requests per parser
parser_requests_error_per_parser{service_name="pender",parser_name="instagram"} 1.0

# TYPE parser_requests counter
# HELP parser_requests Count parsing requests - broken by labels
parser_requests{service_name="pender",parser_name="youtube",parsed_host="www.youtube.com",parsing_status="success"} 1.0
parser_requests{service_name="pender",parser_name="page",parsed_host="sliding.toys",parsing_status="success"} 1.0
parser_requests{service_name="pender",parser_name="page",parsed_host="checkbox.toys",parsing_status="success"} 1.0
parser_requests{service_name="pender",parser_name="instagram",parsed_host="www.instagram.com",parsing_status="error"} 1.0
```

## Checklist

### DIFF
--- a/app/models/media.rb
+++ b/app/models/media.rb
@@ -86,7 +86,7 @@ class Media
     end
     archive_if_conditions_are_met(options, id, cache)
     Metrics.schedule_fetching_metrics_from_facebook(self.data, self.url, ApiKey.current&.id)
-    MetricsService.increment_counter(:media_request_total, labels: { parser: data[:provider], host: URI(data[:url]).host, error: data[:error].nil? ? 'none' : data[:error] })
+    MetricsService.increment_counter(:parsing_requests_total, labels: { parser_name: data[:provider], parsed_host: URI(data[:url]).host, parsing_status: data[:error].nil? ?  'success' : 'error' })
     cache.read(id, :json) || cleanup_data_encoding(data)
   end
 
--- a/app/models/media.rb
+++ b/app/models/media.rb
@@ -86,7 +86,7 @@ class Media
     end
     archive_if_conditions_are_met(options, id, cache)
     Metrics.schedule_fetching_metrics_from_facebook(self.data, self.url, ApiKey.current&.id)
-    MetricsService.increment_counter(:media_request_total, labels: { service: 'pender', parser: self.data['provider'] })
+    MetricsService.increment_counter(:media_request_total, labels: { parser: data[:provider] })
     cache.read(id, :json) || cleanup_data_encoding(data)
   end
 
--- a/app/models/media.rb
+++ b/app/models/media.rb
@@ -85,9 +85,14 @@ class Media
       self.upload_images
     end
     archive_if_conditions_are_met(options, id, cache)
-    Metrics.schedule_fetching_metrics_from_facebook(self.data, self.url, ApiKey.current&.id)
-    MetricsService.increment_counter(:parsing_requests_total, labels: { parser_name: data[:provider], parsed_host: URI(data[:url]).host, parsing_status: data[:error].nil? ?  'success' : 'error' })
-    cache.read(id, :json) || cleanup_data_encoding(data)
+    # Metrics.schedule_fetching_metrics_from_facebook(self.data, self.url, ApiKey.current&.id)
+    MetricsService.increment_counter(:parser_requests_total, labels: { parser_name: data[:provider], parsed_host: URI(data[:url]).host, parsing_status: data[:error].nil? ?  'success' : 'error' })
+    if data[:error].nil?
+      MetricsService.increment_counter(:parser_requests_success, labels: { parser_name: data[:provider], parsed_host: URI(data[:url]).host })
+    else
+      MetricsService.increment_counter(:parser_requests_error, labels: { parser_name: data[:provider], parsed_host: URI(data[:url]).host })
+    end
+      cache.read(id, :json) || cleanup_data_encoding(data)
   end
 
   PARSERS = [
--- a/app/models/media.rb
+++ b/app/models/media.rb
@@ -85,14 +85,9 @@ class Media
       self.upload_images
     end
     archive_if_conditions_are_met(options, id, cache)
-    # Metrics.schedule_fetching_metrics_from_facebook(self.data, self.url, ApiKey.current&.id)
-    MetricsService.increment_counter(:parser_requests_total, labels: { parser_name: data[:provider], parsed_host: URI(data[:url]).host, parsing_status: data[:error].nil? ?  'success' : 'error' })
-    if data[:error].nil?
-      MetricsService.increment_counter(:parser_requests_success, labels: { parser_name: data[:provider], parsed_host: URI(data[:url]).host })
-    else
-      MetricsService.increment_counter(:parser_requests_error, labels: { parser_name: data[:provider], parsed_host: URI(data[:url]).host })
-    end
-      cache.read(id, :json) || cleanup_data_encoding(data)
+    Metrics.schedule_fetching_metrics_from_facebook(self.data, self.url, ApiKey.current&.id)
+    parser_requests_metrics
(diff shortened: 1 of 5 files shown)


vasconsaurus:
@caiosba 
We have a kwai failing test, that we can ignore for now, and fix in a different PR.
But there was a test that failed before and now worked that I'm kind of worried about: `rails test test/integration/medias_integration_test.rb:33`

I'm opening this PR for review, but I think we might want to look into that test.

app/models/media.rb
```diff
@@ -326,4 +326,20 @@ def archive_if_conditions_are_met(options, id, cache)
         self.archive(options.delete(:archivers))
     end
   end
+
+  def parser_requests_metrics
+    url = RequestHelper.normalize_url(self.url)
```
caiosba, line 331: Question: Isn't the URL already normalized at this point? In any case, it should be harmless if the normalization function is idempotent.

app/models/media.rb
```diff
@@ -326,4 +326,20 @@ def archive_if_conditions_are_met(options, id, cache)
...
+    MetricsService.increment_counter(:parser_requests_total)
+    MetricsService.increment_counter(:parser_requests_per_parser, labels: { parser_name: data[:provider], parsing_status: data[:error].nil? ?  'success' : 'error' })
+    MetricsService.increment_counter(:parser_requests, labels: { parser_name: data[:provider], parsed_host: URI(url).host, parsing_status: data[:error].nil? ?  'success' : 'error' })
+    if data[:error].nil?
+      MetricsService.increment_counter(:parser_requests_success)
+      MetricsService.increment_counter(:parser_requests_success_per_parser, labels: { parser_name: data[:provider] })
+    else
+      MetricsService.increment_counter(:parser_requests_error)
+      MetricsService.increment_counter(:parser_requests_error_per_parser, labels: { parser_name: data[:provider] })
+    end
```
caiosba, line 342: Nice and clean!

caiosba (approved):
Very clean, Manu. Thanks!

app/models/media.rb
vasconsaurus: I think it should be, but some tests were failing, seeming to indicate it wasn't (which I found weird). Maybe it's an issue on the tests side. Going to double check this one.

dmou (commented):
looks good! question about naming: we're using prometheus as a protocol for retreiving metrics, but they're collected and utilized as opentelemetry metrics. as such we should use opentelemetry naming conventions (https://opentelemetry.io/docs/specs/semconv/general/metrics/, https://opentelemetry.io/docs/specs/semconv/general/attribute-naming/#recommendations-for-application-developers), including using dot-delimited namespaces. 

in general, namespaces are ideal--it's possible we'll have another service that parses something else entirely like files or user-generated content, and it will also need a `parser_requests_success`/`parser.requests.success` for instance. it might seem like not a big deal because the `serivce.name` is sent up as well, but it means any time you want to query for or generate a dashboard for data about a parser you'll have to query for data about all of those parsers, and it's a very rare situation where you'll need to include those in the same dataset. iirc there was something blocking your ability to use namespaces here, what was the issue?

edit: ooh, answered in those docs. `The Name of an OTLP metric MUST be added as the [Prometheus Metric Name](https://prometheus.io/docs/instrumenting/exposition_formats/#comments-help-text-and-type-information)...` not currently doable with prometheus, looks like we can't do anything about it. that said, we should still prefix these with `pender_` :)

vasconsaurus:
> 
> edit: ooh, answered in those docs. `The Name of an OTLP metric MUST be added as the [Prometheus Metric Name](https://prometheus.io/docs/instrumenting/exposition_formats/#comments-help-text-and-type-information)...` not currently doable with prometheus, looks like we can't do anything about it. that said, we should still prefix these with `pender_` :)

derp, I missed your update! Was going to reply with pretty much the same 😅 
I'll add `pender_`.

app/models/media.rb
vasconsaurus: Updating with the answer from our discussion:

> We do normalize the URL for parsing BUT we return the original URL. Me and Jay think that makes sense: the original URL is readable, and it might look weird for the user to send one URL in and get another. But we wanted to confirm with you if our assumptions make sense. So we do need to normalize `data[:url]` so we can get the host for our metrics.

Also, in the comment I refer to `data[:url]` but in the code I ended up using `self.url`, since we didn't have a preference on which one to use. I thought if it doesn't really make a difference we might as well go with the original one we got.

> Is there a reason we should prefer one over the other? Conceptually what makes more sense for us?
> 
> No, I think they are the same thing. But one is “internal” (thus a model attribute) and the other is just the exposed value.

vasconsaurus:
@dmou I updated with the `pender_`, let me know if it looks ok to you.

dmou:
> @dmou I updated with the `pender_`, let me know if it looks ok to you.

yes, thank you!

PR #506, opened 2025-02-07
6000 – Handle ArchiveOrg error:blocked-url

vasconsaurus:
## Description

When an URL can't be captured ArchiveOrg might send one of the following:

```
{
message: This URL is in the Save Page Now service block list and cannot be captured. Please email us at "info@archive.org" if you would like to discuss this more.,
status: error,
status_ext: error:blocked-url
}
```

```
{
message: We’re currently facing some limitations when it comes to archiving this site. We apologize for any inconvenience this might cause and appreciate your understanding. Please email us at "info@archive.org" if you would like to discuss this more.,
status: error,
status_ext: error:blocked-url
}
```

In both cases, it doesn't make sense for us to retry. We can treat this as a custom exception (`Pender::Exception::BlockedUrl`) and just notify Sentry.

References: CV2-6000

## How has this been tested?

I added a test  and tested manually:
- Through swagger, I made a request to a twitter url with archiver set to `archive_org`
- I checked Sidekiq to see if it would retry
- I set development to notify Sentry
- I checked Sentry to see if it was correctly notified

I tried to de-duplicate some code by creating a `custom_exception` method: https://github.com/meedan/pender/blob/e50d1e0a3d676ef68461981ca25a9b47353f3fce/app/models/concerns/media_archive_org_archiver.rb#L33-L50

But I'm not too sure about this one.

## Checklist

### DIFF
--- a/app/models/concerns/media_archive_org_archiver.rb
+++ b/app/models/concerns/media_archive_org_archiver.rb
@@ -29,11 +29,16 @@ module MediaArchiveOrgArchiver
         else
           data = snapshot_data.to_h.merge({ error: { message: "(#{body['status_ext']}) #{body['message']}", code: Lapis::ErrorCodes::const_get('ARCHIVER_ERROR') }})
           Media.notify_webhook_and_update_cache('archive_org', url, data, key_id)
-          if body['message']&.include?('The same snapshot') ||
-             body['status_ext'] == 'error:too-many-daily-captures' ||
-             body['message']&.include?('error:blocked-url')
+
+          if body['message']&.include?('The same snapshot') || body['status_ext'] == 'error:too-many-daily-captures'
             PenderSentry.notify(
               Pender::Exception::TooManyCaptures.new(body["message"]),
+              url: url,
+              response_body: body
+            )
+          elsif body['status_ext'] == 'error:blocked-url'
+            PenderSentry.notify(
+              Pender::Exception::BlockedUrl.new(body["message"]),
               url: url,
               response_body: body
             )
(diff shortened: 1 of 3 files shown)


app/models/concerns/media_archive_org_archiver.rb
```diff
@@ -29,19 +29,26 @@ def send_to_archive_org(url, key_id, _supported = nil)
...
+              )
           else
             raise Pender::Exception::ArchiveOrgError, "(#{body['status_ext']}) #{body['message']}"
           end
         end
       end
     end
 
+    def custom_exception(status_ext, message)
+      status_ext == 'error:blocked-url' ?  Pender::Exception::BlockedUrl.new(message) : Pender::Exception::TooManyCaptures.new(message)
```
caiosba, line 49: I know that this method is only called for two cases, but in general I think it's more consistent and future-proof to return `Pender::Exception::TooManyCaptures` only if the error is `error:too-many-daily-captures`, instead of "any error that is not `error:blocked-url`". Something like:

```ruby
if status_ext == 'error:blocked-url'
  Pender::Exception::BlockedUrl.new(message)
elsif status_ext == 'error:too-many-daily-captures'
  Pender::Exception::TooManyCaptures.new(message)
else
  Pender::Exception::ArchiveOrgError.new(message)
end
```

caiosba (changes requested):
Thanks Manu! Just a couple of suggestions.

app/models/concerns/media_archive_org_archiver.rb
vasconsaurus: One of the cases dor `TooManyCaptures` is `body['message']&.include?('The same snapshot')`. In this case there is no `status_ext`. 

I can refactor so we take that in account, BUT I'm wondering: since @jayjay-w will have to look into the other cases we discussed in the ticket, would it be better to leave some duplication for now, and create a better abstraction then? So, leave it like this: https://github.com/meedan/pender/blob/752137c8cf828b9be30a6ace7c3dcd6b1b291c04/app/models/concerns/media_archive_org_archiver.rb#L33-L47

@caiosba what do you think?

jayjay-w (approved):

PR #542, opened 2025-04-30
6357 – Fix how we parse a generic Facebook watch page

vasconsaurus:
## Description

A facebook video from facebook watch was sometimes redirecting to the main watch page. Meaning we were getting pretty generic content, like:
- url: `https://www.facebook.com/watch/`
- title: `"Discover Popular Videos"`

This caused `This item already exists` to be raised for different items. So, if an url redirects to the main watch page, we want the title and url to be the url from the request.

References: 6357

## How has this been tested?

```ruby
# "should return url as the title and empty description when directed to main watch page"
rails test test/models/parser/facebook_item_test.rb:568

# "should return title and description when directed to a watch item"
rails test test/models/parser/facebook_item_test.rb:568
```

## Something to think about

I think we might want to change `set_facebook_privacy_error` method name and its error message, or at least the error message. 

https://github.com/meedan/pender/blob/903c7f244c8f5cec748c4c92f11f68ec37fa0c10/app/models/concerns/provider_facebook.rb#L84-L103

### Why?
Because this method doesn't only set the privacy error, but it checks if a page is unavailable, and then sets the error. The reason why a page is unavailable isn't always because of a `privacy_error` or `login_required`. Right now, we have two situations, consent page and watch page, where a page is unavailable because we were redirected to a dead end.

## Checklist

### DIFF
--- a/app/models/concerns/provider_facebook.rb
+++ b/app/models/concerns/provider_facebook.rb
@@ -11,7 +11,7 @@ module ProviderFacebook
         { pattern: /^https:\/\/([^\.]+\.)?facebook.com\/login/, reason: :login_page },
         { pattern: /^https:\/\/([^\.]+\.)?facebook.com\/?$/, reason: :login_page },
         { pattern: /^https:\/\/([^\.]+\.)?facebook.com\/cookie\/consent-page/, reason: :consent_page },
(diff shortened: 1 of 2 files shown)


vasconsaurus:
I'll merge this PR and work on updating `set_facebook_privacy_error` in a new one.

PR #572, opened 2025-08-04
skip_get_canonical_url when the host is instagram

vasconsaurus:
## Description
When we try to get the canonical URL for an Instagram media in production
environment, it always results in a redirection loop. This error happens
in OpenURI, so there is not a lot we can do to fix it.

The error does not affect the functionality of the application,
but it causes unnecessary processing and is really noisy (since we get a
notification in Sentry every time).

Since the parsing of instagram links works fine without the canonical URL,
I'm suggesting we skip this step, but log when we do, if we need to review
at some point.

References: CV2-6210

## How has this been tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can verify the changes. Please describe whether or not you implemented automated tests.

Please describe parts of the change that require extra attention during code review, for example:

- File FFFF, line LL: This refactoring does this and this. Is it consistent with how it’s implemented elsewhere?
- Etc.

## Checklist

### DIFF
--- a/app/models/media.rb
+++ b/app/models/media.rb
@@ -201,6 +201,8 @@ class Media
   # Parse the page and set it to media `doc`. If the `doc` has a tag (`og:url`, `twitter:url`, `rel='canonical`) with a different url, the media `url` is updated with the url found, the page is parsed and the media `doc` is updated
 
   def get_canonical_url
+    return nil if skip_get_canonical_url?
+
     self.doc = self.get_html(RequestHelper.html_options(self.url))
     tag = self.doc&.at_css("meta[property='og:url']") || self.doc&.at_css("meta[property='twitter:url']") || self.doc&.at_css("link[rel='canonical']")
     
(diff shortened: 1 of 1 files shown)


app/models/media.rb
```diff
@@ -210,6 +212,17 @@ def get_canonical_url
     nil # Return nil, since url is empty/not a string
   end
 
+  def skip_get_canonical_url?
+    skip_hosts = ['instagram']
+    host = RequestHelper.parse_url(self.original_url).host
+    if host && skip_hosts.any? { |skip_host| host.include?(skip_host) }
```
caiosba, line 218: That's a bit too broad. It would match things like `instagram-clone.net`, right? Couldn't we move this logic to the parser classes? The default could be `false`, and then the Instagram parser could implement it as `true`.

vasconsaurus: True! I'll look into making it less generic. 

I think we should move this logic into the parser classes, BUT I'm not sure if that is something that we could do right now, at this point we are still initializing Media. This is something we could look into in a separate ticket.

Something else we should look into is we are parsing the url multiple times, I think more times than we need. We could try to re-write it to parse it once and reuse the result where possible.

PR #591, opened 2025-09-10
Update test environment logging settings

vasconsaurus:
## Description

Specially when testing individual tests, it's useful to have more detailed logs.
But we don't need those in CI, and probably don't need them all the time in local testing either.


vasconsaurus:
CodeClimate step failed.

